PR #45, opened 2024-01-05
News

lintangsutawika:
Sites included. Source: https://opennewswire.org/feed/

CC BY

-  360info
-  Africa is a Country
-  Alt News
-  Balkan Diskurs
-  Factly
-  Freedom of the Press Foundation
-  Agenzia Fides
-  Global Voices
-  Meduza
-  Mekong Eye
-  Milwaukee Neighborhood News Service
-  Minority Africa
-  New Canadian Media
-  SciDev.Net
-  The Solutions Journalism Exchange
-  Tasnim News Agency
-  ZimFact

CC BY-SA
-  Liberty TV
-  Oxpeckers
-  Propastop
-  The Public Record

Public Domain
- Caravanserai

 Voice of America intentionally left out so that it does not overlap with MOT.


StellaAthena:
Am I correct in assuming that this is related to https://github.com/r-three/licensed-pile/issues/44? Can you list in the PR which sources you have included / which you plan on doing?

lintangsutawika:
Yes it's related. I'm planning to get through all the website mentioned. Currently have Propublica and Democracy Now.

news/get_text.py
```diff
@@ -0,0 +1,125 @@
...
+    # TODO Save HTML files
+    # Then process/extract
+    get_record_fn = partial(get_record, input_dir=input_dir, date=date, tag=args.tag, attrs=args.attrs)
+    num_workers = mp.cpu_count() if args.num_workers is None else args.num_workers
+    if num_workers == 1:
+        page_data = list(map(get_record_fn, tqdm(page_index)))
+    else:
+        with mp.Pool(num_workers) as p:
+            page_data = []
+            for data in p.map(get_record_fn, tqdm(page_index)):
```
blester125, line 111: I think this can be replaced with `list(p.map(get_record_fn, tqdm(page_index))` instead of manually looping through the results.

Also I don't think you even need to have the `list` call, just `page_data = p.map(...)` as you aren't using `imap`

lintangsutawika: Thanks. Adjusted.

StellaAthena:
Can you add the (estimated) number of documents and number of tokens to the README?

lintangsutawika:
What would be the best way to estimate number of tokens? Would sampling a number of pages and calculating the tokens from there be valid?

StellaAthena:
> What would be the best way to estimate number of tokens? Would sampling a number of pages and calculating the tokens from there be valid?

Yup!

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+def build_url_index(base_url, keyword=None):
+
+    if keyword is None:
+        keyword = [""]
+    elif isinstance(keyword, str):
+        keyword = [keyword]
+
+    tree = sitemap_tree_for_homepage(base_url)
+    page_index = [page.url for key in keyword for idx, page in enumerate(tree.all_pages()) if key in page.url]
```
blester125, line 14: What is the `idx` and enumeration for here? It doesn't seem to be used/saved anywhere

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    if not text:
+        return False
+    elif text.startswith("Attention Required! | Cloudflare"):
+        return False
+    elif text.startswith("403 Forbidden"):
+        return False
+    else:
+        return True
+
+def sanitize_url(url: str) -> str:
```
blester125, line 127: Can we rename this to something like `url_to_filename`? Based on the name I thought it was about url encoding things.

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    base = (
+        url.replace("?", "_")
+        .replace(",", "_")
+        .replace("=", "_")
+        .replace("https://www.", "")
+        .replace("http://www.", "")
+        .replace("https://", "")
+        .replace("/", "_")
+    )
+    return re.sub(r"\s+", "_", base)
```
blester125, line 138: Newline seems to be missing?

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    Simple check to eliminate and filter obviously bad text in paragraph tags.
+    """
+    text = text.strip()
+    text = " ".join(text.split())
+    if not text:
+        return False
+    elif text.startswith("Attention Required! | Cloudflare"):
+        return False
+    elif text.startswith("403 Forbidden"):
```
blester125, line 122: Can this be checked based on the `request.get` status code instead of by parsing text?

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    tree = sitemap_tree_for_homepage(base_url)
+    page_index = [page.url for key in keyword for idx, page in enumerate(tree.all_pages()) if key in page.url]
+    return page_index
+
+
+def get_text_from_page(url=None, html_path=None, tag="article", attrs=None):
+
+    assert bool(url is not None) != bool(html_path is not None)
+
+    if url is not None:
```
blester125, line 22: I think it would be a lot simpler if this function was just html -> text instead of having the ability to pull the page.

Thoughts on if you should pass in raw html strings or the Soup? The former makes it easier to use, the ladder makes it easier to customize the parsing. Unsure what is better.

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+        # headers = {'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36'}
+        headers = {'User-Agent': 'My User Agent 1.0'}
+        page = requests.get(url, headers=headers)
+        soup = BeautifulSoup(page.content, "html.parser")
+    elif html_path is not None:
+        with open(html_path, "rb") as fp:
+            soup = BeautifulSoup(fp, "html.parser")
+
+    text = [soup.title.get_text() if soup.title else ""]
```
blester125, line 31: Does a page not having a title mean it'll start with an extra newline? If so we should fix that, just in case it never gets stripped out later.

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Search for author
+    if byline := soup.find(class_=re.compile("byline")):
```
blester125, line 34: These all seem to do a re search and then have the same code to processes the result, can it be simplified with something like:

```python
if byline  := soup.find(class_=re.compile(r"(byline|post-author|posted-by|..."):
  ...
```

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+        text.append(byline.get_text().strip())
+    elif byline := soup.find(class_=re.compile("post-author")):
```
blester125, line 36: I confirmed it myself, but can you add a comment like

```
# Calling `re.compile(...)` repeatedly in a this processing function (once for each document) is ok because the resulting compiled re object is cached and reused
```

Otherwise I'm going to forget and wonder if repeated calls is an issue in the future lol.

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+        text.append(byline.get_text().strip())
+    elif byline := soup.find(class_=re.compile("posted-by")):
+        text.append(byline.get_text().strip())
+    elif byline := soup.find(class_=re.compile("article__source")):
+        text.append(byline.get_text().strip())
+    elif byline := soup.find(class_=re.compile("author")):
+        text.append(byline.get_text().strip())
+
+    # Search for dateline
+    if dateline := soup.find("time", class_=re.compile("title")):
```
blester125, line 46: Same simplification question as above

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+            # split_p = p.get_text().split("\n")
+            split_p = []
+            text_pieces = []
+            for child in p.children:
+                if type(child) is NavigableString:
+                    text_pieces.extend(child.split("\n"))
+                elif child.name == "br":
+                    split_p.append("".join(text_pieces))
+                    text_pieces = []
+                elif child.name == "em":
```
blester125, line 88: Can the rest of this conditional can be simplified to something like the following?

```python
...
elif child.name in ("em", "a", ...):
  text_pieces.extend(child.get_text())
```

And maybe the `in` check should be against a variable like `FORMATTED_STRING_TAGS` or something?

news/utils.py
```diff
@@ -0,0 +1,138 @@
...
+                elif child.name == "span":
+                    text_pieces.extend(child.get_text())
+
+            # Remaining pieces
+            if text_pieces:
+                split_p.append("".join(text_pieces))
+            text_article = [
+                article_paragraph
+                for s in split_p
+                if is_valid(article_paragraph := s.strip()) and s.strip()
```
blester125, line 105: I think you can avoid the double strip here with

```python
...
if is_valid(article_paragraph := s.strip()) and article_paragraph
...
```

Or you could push the empty check into the `is_valid` function

blester125 (changes requested):
Hi! Sorry for the slow review, feel free to ping me in the future!

Thanks for the PR, it handles a lot!

It does feel unfortunate to need a script that enumerates all the pages and their licenses, but I don't have a great idea on how to do that better.

Right now, the conversion to the dolma formatting and the parsing of the html to raw text is happening at once, do you think it would make sense break those up? have a raw dolma dataset that has the html pages and then the text parsing is a secondary job that uses dolma parallelization? I guess it depends on if we think the html parsing is going to change

Do you have some example outputs you could share?

Again, thanks for you work on this!

lintangsutawika:
Sorry for the delay. Will get to  it this week.

craffel:
@lintangsutawika any bandwidth to finish this up?

blester125:
I made all the changes I had talked about in this branch, and I separated the steps to make the code a bit similar in this PR https://github.com/r-three/licensed-pile/pull/68. We should be able to merge that one

blester125:
Development moved to this branch for easier colab https://github.com/r-three/licensed-pile/pull/68

PR #51, opened 2024-01-15
Feat/wiki scraper

blester125:
This PR adds scripts that can be used to get an xml export of mediawiki sites that don't provide dumps. The resulting dump will contain a list of `<page>`, one for each exported page. Each page has multiple `<revision>` which can be used to create an author list. The most recent `<revision>`'s `<text>` can be used to get the mediawiki markup representation of the page to use as the document text.

An index of pages is built using the `Special:AllPages` query url and then exports are made using `Special:Export`.


wikiscrape/export_pages.py
```diff
@@ -0,0 +1,101 @@
...
+
+parser = argparse.ArgumentParser(description="Export mediawikis as XML")
+parser.add_argument("--wiki", required=True, help="The wiki url we are exporting.")
+parser.add_argument(
+    "--pages",
+    action="append",
+    help="A list of files of pages to export, or a dir to export all. defaults to data/${wiki_name}/pages/.",
+)
+# Using firefox I didn't have issues sending a lot of pages at once, but I was
+# getting URI too long errors when using requests.
```
craffel, line 30: Should we specify a user-agent etc.?

wikiscrape/export_pages.py
```diff
@@ -0,0 +1,101 @@
...
+    "--page_limit", default=35, help="The max number of pages to export at once."
+)
+parser.add_argument(
+    "--output_dir",
+    help="Where to save the xml export. defaults to data/${wiki_name}/export/.",
+)
+# TODO: Implement this if we find a wiki that has this enabled.
+parser.add_argument(
+    "--listauthors",
+    help="Use the listauthors url param instead of getting multiple revisions. UNIMPLEMENTED.",
```
craffel, line 41: If you want to be fancy, raise a NotImplementedError if this is set?

wikiscrape/list_pages.py
```diff
@@ -0,0 +1,122 @@
...
+        for link in links.find_all("a"):
+            pages.append(
+                urllib.parse.unquote(removeprefix(link.attrs["href"], "/wiki/"))
+            )
+    logging.info(f"Found {len(pages) - page_count} pages")
+
+    # Find a pagenation link
+    if nav := soup.find("div", {"class": "mw-allpages-nav"}):
+        for link in nav.find_all("a"):
+            # Pagenation links look like "Next Page (${page title})"
```
craffel, line 71: It's pagination fwiw

craffel (commented):
Which wikis have you tried this on? Can it be easily parallelized to get more than 35 pages at once?

StellaAthena:
I think it's a good idea to replace Wikipedia's custom mathematics syntax with LaTeX. Does it make sense to do it at this stage of the pipeline, or later?

wikiscrape/export_pages.py
blester125: Added

wikiscrape/README.md
```diff
@@ -0,0 +1,19 @@
...
+1. Find all the namespaces that pages are listed under with `python get_namespaces.py --wiki ${wiki_url}`. This saves a mapping of namespace names to id's in `data/${wiki_name}/namespaces.json`.
+2. Get all the pages under each namespace by following pagenation links using `python list_pages.py --wiki ${wiki_url} -ns 0 -ns 1...`. The namespaces we want to scrape are generally:
+  * `(Main)`: 0
+  * `Talk`: 1
+  * `UserTalk`: 3
+Either the integer or the name can be used as input. This generates lists of page titles at `data/${wiki_name}/pages/${ns}.txt`.
+3. Get the XML export of these pages with `python export_pages.py --wiki ${wiki_url}`. This get xml exports of the all the pages exported pages. It currently fetches all revisions so that we can build a complete author list. This will create a sharded xml export at `data/${wiki_name}/export/${shard_idx}-pages.xml`. The `<text>` tag contains the wikimedia markup.
+
+
+**TODO:** Is this exported format the exact same as the published mediawiki dumps to the point we can reuse code?
```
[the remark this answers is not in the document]
blester125: Yeah I can post an example dump

wikiscrape/list_pages.py
blester125: Fixed

wikiscrape/export_pages.py
blester125: Do you mean we should pretend to be firefox in case they have some checks for automatic tools?

blester125:
@craffel Yeah it should be easy to parallelize this, it runs off files which list page titles (one per line) so you can parallelize over the files (and we are already parallelizing over wiki's) and we can split the inputs pretty easily for more parallelism.

@StellaAthena I think it would best to have that happen later. I was thinking after this export there would be a step that converts from these xml to dolma which would have raw wiki markup as the `text` field. Then the next step would be converting wikitext to plaintext and that is where the math conversion would happen.

wikiscrape/export_pages.py
craffel: Yes.

blester125: What about something like this?
```
Mozilla/5.0 (compatible; Licensed-Pile-bot/0.1; +http://www.github.com/r-three/licensed-pile)
```

Would get some goodwill from site owners if we identify as a bot and have a link with info about why we are scraping them?

craffel: Sounds good 👍

blester125:
Datasets have been uploaded to https://huggingface.co/datasets/blester125/wiki-dolma

WikiMedia + Talk pages are cleaner and have 14.6 Billion tokens.
WikiTeam3 wikis have 65.1 Billion tokens. They are less clean, various default/boilerplate pages pop up

PR #52, opened 2024-01-18
Update StackExchange Preprocessing.

blester125:
This PR updates how stack exchange preprocessing can be done.

It adds two new flags, `--sort` lets you control how answers are added to the document. `--sort=time` orders the answers by the date they were first posted. `--sort=votes` orders the answers by the score they get, with the "accepted answer" being the first in the list.

This PR also adds a `--skip_comments` flag, which allows one to skip adding the comments to the generated document as stackexchange seems to consider comment ephemeral.

It also adds a new metadata field which is the set of all licenses that appear on the comments/answers/the question that go into a single document. Then comments/answers are posted way after the original question, the version of the CC license can change.

These changes are based on discussions from the last meeting.

### DIFF
--- a/stackexchange/preprocess.py
+++ b/stackexchange/preprocess.py
@@ -13,7 +13,7 @@ import os
 import shelve
 import urllib.parse
 from dataclasses import dataclass
-from typing import List
+from typing import Dict, List, Sequence
 
 import bs4
 import tqdm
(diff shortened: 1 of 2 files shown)


nkandpa2 (approved):

stackexchange/preprocess.py
```diff
@@ -133,12 +153,14 @@ def process_comment(comment):
...
       The text of the comment
       The date the comment as created
+      The license that applies to the comment
     """
     return (
         get_attr(comment, "PostId"),
         get_attr(comment, "UserId"),
         get_markdown_text(comment),
         get_date(get_attr(comment, "CreationDate")),
+        stackexchange_license(get_attr(comment, "ContentLicense")),
```
craffel, line 163: Is this not always the same for all SE sites?

craffel (approved):

craffel:
Thanks!!

PR #56, opened 2024-02-06
PubMed Central

alon-albalak:
This PR contains the initial code for downloading/extracting/converting PubMed Central articles.

It introduces 3 main files:

- `get-filelist.sh` (downloads the metadata for all pubmed articles and filters out data with non-permissive licenses)
- `download-convert-to-md.sh` (downloads individual pubmed articles as nxml files and converts them to md)
- `to-dolma.py` (formats the articles in the dolma style)

To test the code without running over the full dataset, change line 4 in `download-convert-to-md.sh` to `python3 download_and_convert_to_md.py --filelist data/permissive_filelist.txt --total_docs 1000`


alon-albalak:
Thanks for taking a look @craffel ! I modified the download-convert-to-md.sh script as you requested.

For some example .md files, check out `PMC553995.md` and `PMC544568.md` for tables that were formatted nicely (you'll have to look at the raw files because github hides the table).
See `PMC509239.md` for a table that is pretty poorly formatted because it is meant to have many figures inside the table. Tables 1 and 2 look pretty rough, but Tables 3 and 4 came out okay.
See `PMC1149498.md` for a fairly complex equation that couldn't be formatted as native markdown, and was left as LaTeX, and the modelling section of `PMC544568.md` for some equations formatted in html/markdown. 


As a side note, some files also include supplemental tables as attachments which I didn't integrate into the text because they are pretty infrequent (only 13 out of 10000 files have a supplemental table) and very large when they do exist (on average 1.5M characters). We definitely could include them inline but not sure it would be useful.

craffel:
Huh, yeah, those seem pretty reasonable. It's a little weird that it includes image tags - I'm not sure that's something that would be helpful to model, maybe we should try to strip them out? Also, regarding the markdown formatting, I think roughly speaking we should aim to have umarked text. Is there a nice pipeline to go from markdown to unmarked text (with the exception of, perhaps, equations and tables)? Thanks.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+parser.add_argument("--filelist", help="The path to the filelist.txt file.")
+parser.add_argument(
+    "--output_dir", default="data/raw/", help="Where the markdown files go."
+)
+parser.add_argument("--total_docs", default=0, type=int, help="Total number of documents to convert, for debugging.")
+
+BASE_URL = "https://ftp.ncbi.nlm.nih.gov/pub/pmc/"
+
+
+def download(f_url: str, output_dir: str):
```
blester125, line 21: Is there a reason to do a subprocess call to `wget` over making the request from python? If something like the code from https://github.com/r-three/licensed-pile/blob/d2dadf0de22c832459be2fc0e11bc2387b625e34/wikiscrape/utils.py#L20 will work for you we can move that into a shared location for reuse.

Can you test if that snippet works for you?

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+            # extract nxml file
+            tar.extract(nxml)
+
+        # convert nxml to markdown
+        pmcid = nxml.split("/")[0]
+        os.system(f"pandoc --quiet -f jats {nxml} -o {pmcid}.md --wrap=none")
+
+        # remove extracted files
+        os.system(f"mv {pmcid}.md {output_dir} && rm -r {nxml.split('/')[0]}")
```
blester125, line 53: python has built in functionality to do things like move files, e.g. `os.rename`, we should try to use those instead of subprocess calls.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+    except:
+        print(f"Error extracting {t}")
+        import traceback
```
blester125, line 57: We should have our imports at the top of the file, imports are cheap (unless some other developer did something very wrong lol) so it's fine to always import things, even if they aren't needed on all code paths.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+import tarfile
+
+from tqdm import tqdm
+
+cur_dir = os.getcwd()
+
+parser = argparse.ArgumentParser(description="Convert xml documents to markdown.")
+parser.add_argument("--filelist", help="The path to the filelist.txt file.")
+parser.add_argument(
+    "--output_dir", default="data/raw/", help="Where the markdown files go."
```
blester125, line 14: Nit (meaning it isn't a huge deal if it doesn't get changed): These files are getting converted to markdown by pandoc so they aren't raw anymore, maybe change the dir to `data/md`?

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+def extract_and_convert_tarball(t: str, output_dir: str):
+    if not os.path.exists(t):
+        return
+    try:
+        with tarfile.open(t) as tar:
+            files = tar.getnames()
+            nxml = [f for f in files if f.endswith(".nxml")]
+
+            # make sure there's only one nxml file
```
blester125, line 41: How often is there more that one xml file? Does it mean their data is messed up or is it an edgecase we don't handle? If the latter, is it worth handling?

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+        traceback.print_exc()
+
+
+def extract_and_convert_tarball(t: str, output_dir: str):
+    if not os.path.exists(t):
+        return
+    try:
+        with tarfile.open(t) as tar:
+            files = tar.getnames()
```
blester125, line 38: It doesn't look like `files` is reused, can it be rolled into the list comprehension?

```python
nxml = [f for f in tar.getnames() if f.endswith(".nxml")]
```

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+            # make sure there's only one nxml file
+            assert len(nxml) == 1
+            nxml = nxml[0]
+
+            # extract nxml file
+            tar.extract(nxml)
+
+        # convert nxml to markdown
+        pmcid = nxml.split("/")[0]
+        os.system(f"pandoc --quiet -f jats {nxml} -o {pmcid}.md --wrap=none")
```
blester125, line 50: Can you add comments about what the flags for pandoc do? Like `-f jats`?

blester125, line 50: The recommended way to run subprocesses in python is with the `subprocess.run` function over `os.system`. Can we switch to that?

https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # remove extracted files
+        os.system(f"mv {pmcid}.md {output_dir} && rm -r {nxml.split('/')[0]}")
+
+    except:
```
blester125, line 55: It looks like there can be error beyond extraction that end up in this `except:` can you try to refactor a bit to have different excepts for things like failed to extract, failed to convert with pandoc, etc?

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+        print(f"Error extracting {t}")
```
blester125, line 56: Can we switch to using the logging module instead of printing for error messages like these?

Same for other cases where print is used.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+        import traceback
+
+        traceback.print_exc()
+        import time
+
+        time.sleep(0.1)
```
blester125, line 62: Why is a sleep needed here?

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def download_and_convert(line: str, output_dir: str):
+    # split line into parts
+    partial_path, journal, PMCID, PMID, license = line.split("\t")
```
blester125, line 67: It doesn't look like anything beyond partial path is used, so you could probably just index this with `[0]`. Either way, we should try to reserve all uppercase names for global variables instead of locals like `PMCID`

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # create paths for the url and the destination of the markdown file
+    f_url = os.path.join(BASE_URL, partial_path)
```
blester125, line 70: Can we move `BASE_URL` to a default parameter value? Just in case it ever needs to be changed.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+    f_dest = os.path.join(output_dir, partial_path.split("/")[-1])
+
+    try:
+        download(f_url, output_dir)
+        extract_and_convert_tarball(f_dest, output_dir)
+
+        # delete the tarball
+        os.system(f"rm {f_dest}")
```
blester125, line 78: Something like [shutil.rmtree](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) is probably a better way to delete the tarballs instead of a subprocess call.

We can probably skip deleting these all together. The current code is uses `-nc` when downloading so if we ever need to re-run processing with updated code we could skip downloading if we don't delete the tarball.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+    with open(filelist) as fh:
+        lines = fh.read().split("\n")
+
+    # ignore the header
+    lines = lines[1:]
+
+    if args.total_docs > 0:
+        lines = lines[:args.total_docs]
+
+    p = mp.Pool(64)
```
blester125, line 98: We should have the number of processes as a CLI flag, probably with a default to `mp.cpu_count`

blester125, line 98: We should use a context manager instead of opening up the pool directly. Also tqdm is able to wrap iterators so you don't have to all `.update`.

Does a code snippet like this work?

```python
with mp.Pool(args.processes) as p:
  # Use list to force the execution of the imap iterable within the context of the multiprocessing pool.
  _ = list(tqdm(p.imap(functools.partial(download_and_convert, output_dir=args.output_dir), lines, total=len(lines))))
```

pubmedcentral/get-filelist.sh
```diff
@@ -0,0 +1,14 @@
...
+
+NONPERMISSIVE_LICENSES="NO-CC CODE\|CC BY-NC\|CC BY-ND\|CC BY-NC-ND\|CC BY-NC-NA\|CC BY-NC-SA"
+
+# download full filelist
+# -nc: no clobber, do not overwrite existing files
+wget -nc https://ftp.ncbi.nlm.nih.gov/pub/pmc/oa_file_list.txt -P data/
+
+PERMISSIVE_FILELIST_DEST="data/permissive_filelist.txt"
+# -v invert match (find lines without these patterns), -w match whole word
+grep -vw "${NONPERMISSIVE_LICENSES}" data/oa_file_list.txt > ${PERMISSIVE_FILELIST_DEST}
```
blester125, line 11: The second interpolation isn't wrapped in `""` like the first one is.

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+from licensed_pile.write import to_dolma
+
+parser = argparse.ArgumentParser(description="Collect PubMedCentral into Dolma format.")
+parser.add_argument("--filelist", help="The path to the filelist.txt file.")
+parser.add_argument(
+    "--data_dir", default="data/raw/", help="Path to the directory of markdown files."
+)
+parser.add_argument(
+    "--output_dir",
+    default="data/dolma/",
```
blester125, line 16: Other datasources have been using `data/${source_name}` as the output dir, we should follow this.

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+    file, citation, accessionID, PMID, lic = example.split("\t")
+    file = file.split("/")[-1].replace("tar.gz", "md")
+    with open(os.path.join(data_dir, file)) as f:
+        text = f.read()
+
+    return {
+        "id": accessionID,
+        "text": text,
+        "source": source_name,
+        "added": datetime.datetime.utcnow().isoformat(),
```
blester125, line 47: Does the xml version have the creation date in it's metadata? If so is there a way to access that here to include as the (optional) `created` field dolma supports?

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+        "metadata": {
+            "license": str(LICENSE_MAP[lic]),
+            "url": f"https://www.ncbi.nlm.nih.gov/pmc/articles/{accessionID}/",
```
blester125, line 50: Can we replace this with a `BASE_URL` variable like in the other script?

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+LICENSE_MAP = {
+    "CC0": PermissiveLicenses.CC0,
+    "CC BY": PermissiveLicenses.CC_BY,
+    "CC BY-SA": PermissiveLicenses.CC_BY_SA,
+}
+SOURCE_NAME = "PubMed Central"
+
+
+def format_dolma(example: str, data_dir: str, source_name: str = SOURCE_NAME):
+    file, citation, accessionID, PMID, lic = example.split("\t")
```
blester125, line 38: Given it's use later, `journal` seems like a better variable name than `citation`

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+    file = file.split("/")[-1].replace("tar.gz", "md")
```
blester125, line 39: Using they python path manipulation tools like `os.path.basename` is preferable to doing manual split + index like this, unless you are using multiple parts of the path.

pubmedcentral/to-dolma.py
blester125, line 38: What is the difference between `accessionID` and `PMID`? Should we save the `PMID` into the metadata field? If we aren't can we replace it with `_,` as we aren't using it?

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+    with open(os.path.join(data_dir, file)) as f:
+        text = f.read()
+
+    return {
+        "id": accessionID,
+        "text": text,
+        "source": source_name,
+        "added": datetime.datetime.utcnow().isoformat(),
+        "metadata": {
```
blester125, line 48: Would it be easy to add author information to the metadata? We have that for several other sources

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+def main(args):
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    with open(args.filelist) as f:
+        examples = f.read().split("\n")
+
+    # Remove the header
+    examples = examples[1:]
+
+    for example in examples:
```
blester125, line 65: Why are we printing all the lines?

pubmedcentral/to-dolma.py
```diff
@@ -0,0 +1,74 @@
...
+            "journal": citation,
+        },
+    }
+
+
+def main(args):
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    with open(args.filelist) as f:
+        examples = f.read().split("\n")
```
blester125, line 60: Similar to the other script, is `files` a better variable name here?

blester125 (changes requested):
This is a great start thanks!

I made some comments that should helpfully help the reliability of code and reduce the number of subprocess calls.

Did you explore using pandoc as part of a dolma parallel processor? IIUC dolma has functionality to automatically parallelize/get the most out of hardware/scale to multiple worker settings. As it is now, if the pandoc part gets to be a bottleneck it seems like it could be hard optimize.

blester125:
> Huh, yeah, those seem pretty reasonable. It's a little weird that it includes image tags - I'm not sure that's something that would be helpful to model, maybe we should try to strip them out? Also, regarding the markdown formatting, I think roughly speaking we should aim to have umarked text. Is there a nice pipeline to go from markdown to unmarked text (with the exception of, perhaps, equations and tables)? Thanks.

I think things like this would probably be best implemented as a second step in the pipeline, one that uses the dolma parallel processors, wholes whole job would be converting the markdown into an even plainer text format.

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,109 @@
...
+from tqdm import tqdm
+
+cur_dir = os.getcwd()
+
+parser = argparse.ArgumentParser(description="Convert xml documents to markdown.")
+parser.add_argument("--filelist", help="The path to the filelist.txt file.")
+parser.add_argument(
+    "--output_dir", default="data/raw/", help="Where the markdown files go."
+)
+parser.add_argument("--total_docs", default=0, type=int, help="Total number of documents to convert, for debugging.")
```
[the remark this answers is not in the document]
alon-albalak: It's possible, but I don't think it simplifies the code. This script is called from ```pubmedcentral/download-convert-to-md.sh``` which sets the default value to 0.

If we change args.total_docs to None by default, then we would still need to set the default argument in ```pubmedcentral/download-convert-to-md.sh``` to be the string None, and then ```pubmedcentral/download_and_convert_to_md.py``` needs to convert the string back to None value anyway

pubmedcentral/download_and_convert_to_md.py
alon-albalak: That works for me, let's get that added to a shared location on the main branch!

pubmedcentral/PMC1149498.md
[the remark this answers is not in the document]
alon-albalak: Done!

pubmedcentral/download_and_convert_to_md.py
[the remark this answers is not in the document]
blester125: Makes sense, None in bash isn't handled well lol

pubmedcentral/download_and_convert_to_md.py
alon-albalak: @blester125 Pinging you about getting your get_page function pushed to the main branch. I'd be happy to create a new pull request just for getting that added. Maybe we can create a utils folder that's shared across domains?

pubmedcentral/to-dolma.py
alon-albalak: Yeah, no worries. Just added it in commit a9009b972cd755a156d420c15b0e2a683fe1b4a5

alon-albalak:
> Did you explore using pandoc as part of a dolma parallel processor? IIUC dolma has functionality to automatically parallelize/get the most out of hardware/scale to multiple worker settings. As it is now, if the pandoc part gets to be a bottleneck it seems like it could be hard optimize.

I did not consider this. Do you have an example of how that would look?

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,179 @@
...
+    authors = []
+
+    tree = ET.parse(nxml_file)
+
+    # search for author tags
+    for author in tree.findall(".//contrib[@contrib-type='author']"):
+        surname = author.find("name/surname")
+        given_names = author.find("name/given-names")
+        if surname is not None and given_names is not None:
+            authors.append(f"{surname.text}, {given_names.text}")
```
blester125, line 50: Given that we have it  split in the metadata lets store the authors as `List[{"first": str, "last": str}]`

pubmedcentral/download_and_convert_to_md.py
```diff
@@ -0,0 +1,179 @@
...
+        # get pmcid
+        pmcid = nxml.split("/")[0]
+
+        # get authors from nxml file
+        authors = get_authors(nxml, pmcid)
+        # write to file
+        with open(
+            f"{os.path.join(args.author_dir, pmcid)}.txt", "w", encoding="utf-8"
+        ) as f:
+            f.write(" and ".join(authors))
```
blester125, line 98: I think this will need to change a bit if we store authors as a list as mentioned above

blester125 (changes requested):
One tiny tweak and then I think it ready to merge! Great Job on this!

alon-albalak:
Updated the author list as `List[{"first": str, "last": str}]`

pubmedcentral/to-dolma.py
alon-albalak: Done!

pubmedcentral/download_and_convert_to_md.py
alon-albalak: Done

pubmedcentral/download_and_convert_to_md.py
alon-albalak: Done

blester125 (approved):

PR #69, opened 2024-05-07
Scape examples from foodista

blester125:
This PR scapes training data from foodista, a shared collection of recipes and information about cooking tools, techniques, and ingredients that is distributed under the CC-BY 3.0 license.

Data is collected in 4 steps:
1. an index of pages is built from the sitemap
2. all pages are downloaded
3. the pages are converted to dolma examples, with the raw html as the `"text"` key.
4. The html is parsed as part of a dolma processor.

closes https://github.com/r-three/licensed-pile/issues/11


food/examples/4.json
```diff
@@ -0,0 +1,21 @@
...
+  "source": "foodista",
+  "added": "2024-05-07T18:51:12.319767",
+  "created": "2007-12-11T00:00:00",
+  "metadata": {
+    "license": "Creative Commons - Attribution - https://creativecommons.org/licenses/by/3.0/",
+    "url": "https://foodista.com/blog/2007/12/11/theres-salt-and-theres-salt",
+    "authors": [
+      [
+        "Gâteau au Foie ...",
```
craffel, line 12: Some of these comments seem to be some kind of truncated version of articles posted elsewhere (probably for backlinks?) Should we try to omit those? The legit comments may be worth keeping.

food/preprocess.py
```diff
@@ -0,0 +1,183 @@
...
+        author = author.get_text().strip()
+        result.append(f"By: {author}")
+        authors.append((author, user_id))
+    else:
+        logger.warning(f"Failed to find author for example: {idx}")
+
+    # Find the date it was published.
+    if date := soup.find("div", class_="pane-node-created"):
+        date = date.get_text().strip()
+        result.append(f"Published: {date}")
```
craffel, line 103: I'm not sure we need the publication date, but if you have a specific reason for including it I'm fine with it.

food/examples/2.json
```diff
@@ -0,0 +1,21 @@
...
+  "source": "foodista",
+  "added": "2024-05-07T19:57:00.094171",
+  "created": "2007-12-10T00:00:00",
+  "metadata": {
+    "license": "Creative Commons - Attribution - https://creativecommons.org/licenses/by/3.0/",
+    "url": "https://foodista.com/blog/2007/12/10/cheese-glorious-cheese",
+    "authors": [
+      [
+        "Ken Wetherell",
+        ""
```
craffel, line 13: Should the author list have more structure? For the purposes of attribution I think we only need the author name as displayed, fwiw, and not the user id or anything.

blester125: I updated it to only output the author names for now. The parser still finds the user ids (and is robust if it fails to) so we can turn it on in the future.

For Stack Exchange our author list includes both the user names and the user link but they are a single list, not grouped like this one was. I don't think it is worth making sure they all match exactly. IIRC, someone said they we technically don't even need specific authors for attribution, just the site it comes from :woman_shrugging:

craffel: Yeah, I'm not sure there is a specific set of practices, but CC has some recommendations that do include crediting the author: https://wiki.creativecommons.org/wiki/Recommended_practices_for_attribution 

The reason the SE parsing includes author profile page links is that SE actually requests even more stuff for attribution: https://stackoverflow.blog/2009/06/25/attribution-required/ so it seemed like a good practice to try to include that. Unless foodista makes a similar request, I think we can avoid including the user page and the name is enough.

PR #70, opened 2024-05-13
USGPO

### LINKED ISSUE #64: US Government Publishing Office
[US GPO](https://www.gpo.gov) is the agency responsible for publishing documents authored by the US federal government (and thus are public domain) and they [provide an API](https://api.govinfo.gov/docs/) for accessing these documents and associated metadata.

nkandpa2:
This PR uses the USGPO developer API to collect documents published by the USGPO. This will close #64.


usgpo/get-links.py
```diff
@@ -0,0 +1,185 @@
...
+            pbar.update(1)
+
+
+def main(args):
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    package_queue = queue.Queue()
+    metadata_queue = queue.Queue()
+
+    with ThreadPoolExecutor(max_workers=args.workers + 2) as executor:
```
blester125, line 164: This parallelism is pretty complicated, it is possible to fetch all the packages in parallel, collect results, and then fetch the metadata from each package in parallel afterwards?

StellaAthena:
What is still "WIP" about this? What do we need to add so it's ready to go?

nkandpa2:
This PR should be ready to go. I've started a run going back to 1990. So far about 70K document links have been collected with at least a handful of documents coming from each year and no failures.

usgpo/download-files.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def construct_record(api_key, file):
+    file_url = file["links"].get("txtLink")
+    if file_url is None:
+        return None
+    raw_html = download_file(api_key, file_url)
+    parsed_text = parse_html(raw_html)
+
+    return {
```
blester125, line 81: The data if formatted incorrectly,

`title` `author` publisher`, `category` should all be in the `metadata` field.
`date` -> `created`

Do we need the `html` field? If we want to give people access to it, it would probably be to change the code so this outputs a dolma file here the html is in the `text` field and the path is `.../raw/documents`. Then have a preprocess script that uses the dolma parallel processors to convert the html to text and save it to `.../v0/documents`

usgpo/download-files.py
```diff
@@ -0,0 +1,115 @@
...
+        "html": raw_html,
+        "text": parsed_text,
+        "source": SOURCE_NAME,
+        "added": datetime.datetime.utcnow().isoformat(),
+        "metadata": {"license": str(PermissiveLicenses.PD), "url": file_url},
+    }
+
+
+def generate_records(args):
+    with jsonlines.open(args.links_file, mode="r") as reader:
```
blester125, line 97: Is there a speed boost from using jsonlines? It seems simple enough to just call `json.loads` on each line.

usgpo/download-files.py
```diff
@@ -0,0 +1,115 @@
...
+        with ThreadPoolExecutor(max_workers=args.workers) as executor:
```
blester125, line 98: Is there a reason to use the ThreadPoolExecutor over `multiprocessing.dummy.Pool`?

usgpo/get-links.py
```diff
@@ -0,0 +1,187 @@
...
+            "PAI",
+            "PLAW",
+            "USCODE",
+        ],
+    )
+    args = parser.parse_args()
+    return args
+
+
+def api_query(endpoint, headers, params):
```
blester125, line 50: This should probably move to a utils file to be shared with the `download-files.py` script.

usgpo/get-links.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+        if response.status_code == 200:
+            output = response.json()
+
+            for record in output["packages"]:
+                package_queue.put(record)
+                pbar.update(1)
+
+            url = output["nextPage"]
+            offset_mark = None
```
blester125, line 100: Can you document this `offset_mark` a bit, being `"*"` the first time and `None` the rest is a bit confusing.

usgpo/download-files.py
```diff
@@ -0,0 +1,115 @@
...
+    pre_tag = soup.find("pre")
+    if pre_tag:
+        parsed_text = pre_tag.get_text()
+    else:
+        parsed_text = text
+    return html.unescape(parsed_text)
+
+
+def construct_record(api_key, file):
+    file_url = file["links"].get("txtLink")
```
blester125, line 75: The naming is confusing here, `links` makes it sound like a list of links but `.get` makes it seems like a dict representing one link?

usgpo/get-links.py
```diff
@@ -0,0 +1,187 @@
...
+        )
+
+        # `args.workers` threads for getting package metadata
+        for _ in range(args.workers):
+            executor.submit(
+                get_package_metadata, args.api_key, package_queue, metadata_queue
+            )
+
+        # One thread for writing out the package metadata to disk
+        executor.submit(write_metadata, args.output_dir, metadata_queue)
```
blester125, line 179: I think you have a race condition in this parallelism. Let the package_queue have 2 packages followed the `None` sentinel, and there are 2 two workers. 1 worker gets the links quickly while the second worker gets a `429` and sleeps for an hour. worker 1 writes the package metadata to the metadata_queue and gets the next item from the package queue, the `None`, and writes it to the metadata queue. This means the writer will see the `None` before the metadata for package 2, the writer will close, and the package 2 metadata will be lost.

blester125: I think the way people generally do this is each worker needs to add a sentinel to the writer queue and the write only stops when it sees #workers sentinels

usgpo/get-links.py
```diff
@@ -0,0 +1,187 @@
...
+        "--start-date",
+        required=True,
+        help="Start date in ISO8601 format (yyyy-MM-dd'T'HH:mm:ss'Z')",
+    )
+    parser.add_argument("--output-dir", required=True, help="Path to output directory")
+    parser.add_argument("--workers", type=int, default=20, help="Number of threads")
+    parser.add_argument(
+        "--collections",
+        nargs="+",
+        default=[
```
blester125, line 28: prefer a tuple for read only things like this.

blester125 (changes requested):
Some small changes and the data format is currently wrong afaict

PR #71, opened 2024-05-13
USPTO

### LINKED ISSUE #9: Patent Data
Domain: Patents

Can we use the Google Patents data for this?

It might be possible to use C4/Common Crawl data for this as `patents.google.com` is one of the most represented domains in c4

baberabb:
This PR adds the code to process the USPTO dataset extracted from [Google Patents Public Dataset](https://cloud.google.com/blog/topics/public-datasets/google-patents-public-datasets-connecting-public-paid-and-private-patent-data) and uploaded to HF. The dataset covers all US patent applications until Oct 27, 2023 (including historical).

closes #9 

Edit: switched over to pandoc as that does the latex formatting pretty well, and also works well with the overall html.


uspto/process_uspto.sh
```diff
@@ -0,0 +1,44 @@
...
+
+
+# Clone MathML to LaTeX converter
+git clone https://github.com/baberabb/mathml-to-latex.git
+
+# Navigate to the Node.js project directory
+cd mathml-to-latex || exit
+
+# Install Node.js dependencies
+npm install
```
blester125, line 23: I think we should segment this into two scripts, one that sets up the mathml-to-latex server (does the clone, install, etc) and one that runs data processing (start the server, run the code, kill the server, etc)

I'll do a more through review later today

uspto/uspto-to-dolma.py
```diff
@@ -0,0 +1,228 @@
...
+    max_concurrency: int = 4,
+) -> Iterable[dict]:
+    """
+    This function `run_dataset` scans a dataset located in a directory, converts each file in the dataset to a desired
+    format using an API endpoint,and returns an iterable of dictionaries containing the converted data.
+
+    Parameters:
+    - `data_dir` (str): The directory where the dataset is located. Default value is "./data/uspto/".
+    - `url` (str): The API endpoint URL for converting the dataset files. Default value is "http://localhost:3000/convert".
+    - `limit` (int): The maximum number of files to convert. Default value is 0, which means convert all files in the dataset.
```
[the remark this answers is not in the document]
baberabb: yeah. good catch!

uspto/uspto-to-dolma.py
```diff
@@ -0,0 +1,228 @@
...
+        data_dir += r"/"
+
+    # columns to use
+
+    file_names = glob.glob(data_dir + r"*.parquet")
+    if limit > 0:
+        limit //= len(file_names)
+        logger.info(f"Processing {limit} entries each from {len(file_names)} files.")
+    args = [(x, url, limit) for x in file_names]
+    with multiprocessing.get_context("spawn").Pool() as pool:
```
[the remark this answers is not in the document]
baberabb: So I'm mainly relying on `polars` here, which defaults to using all available cores. It usually works well enough but the API calls act as a blocker so using multiprocessing to process each file and let polars handle each individual file.

baberabb:
Thanks for the review @blester125! I think i have addressed all your points but let me know if I missed anything! Added the readme as well. The only thing left are the tables. They are mostly fine but the whitespace formatting is quite off. Will bring it up in the meeting!

uspto/uspto-to-dolma.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Example usage:
+    ```python
+    for data in run_dataset(data_dir=r"./data/uspto/", url="http://localhost:3000/convert", limit=10, max_concurrency=2):
+        # Process each converted data entry
+        print(data)
+    ```
+    """
+
+    file_names = glob.glob(os.path.join(data_dir + "*.parquet"))
```
blester125, line 63: Nit: You shouldn't need the os.path.join here right? `data_dir` is typed as a string so it shouldn't do anything?

blester125 (approved):

uspto/uspto-to-dolma.py
blester125: Thanks! Should be good to merge!

baberabb:
> Thanks! Feel free to merge when ready!

Whenever you're ready!

PR #72, opened 2024-05-14
Stackv2

Muennighoff:


Muennighoff:
Still waiting for the dataset to be released (hopefully soon), but does this look more or less reasonable @blester125 ? Is there anything else I can do while the DS is not yet released?

stackv2/to_dolma.py
```diff
@@ -0,0 +1,63 @@
...
+
+from datasets import load_dataset
+
+from licensed_pile.licenses import PermissiveLicenses
+from licensed_pile.logs import configure_logging
+from licensed_pile.write import to_dolma
+
+SOURCE_NAME = "The Stack v2"
+
+logger = configure_logging("court-listener-opinion")
```
blester125, line 13: copypasta lol

stackv2/to_dolma.py
```diff
@@ -0,0 +1,63 @@
...
+            "added": datetime.utcnow().isoformat(),
+            "created": "2024-02-29",
+            "metadata": {
+                "license": str(LICENSE_MAP[x["license"]]),
+                "url": "https://github.com/" + x["repo"] + "/tree/" + x["revision_id"],
+            },
+        }
+    )
+
+    output_file_base_name = os.path.join(args.output_dir, "starcoder2data.jsonl.gz")
```
blester125, line 39: You don't need to join the output_dir and `"startcoder2data.jsonl.gz"`, you should just pass the latter.

stackv2/to_dolma.py
```diff
@@ -0,0 +1,63 @@
...
+    to_dolma(ds, args.output_dir, output_file_base_name, args.shard_size)
+    logger.info(f"Saved {args.data_path} as dolma sharded files at {args.output_dir}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Convert csv data to dolma.")
```
blester125, line 45: convert HF data to dolma

stackv2/sample.json
blester125, line 1: What is this file? Fake data that has the same format as the HF data once loaded? Is there a reason to check it in?

stackv2/to_dolma.py
```diff
@@ -0,0 +1,63 @@
...
+    parser.add_argument(
+        "--data_path",
+        default="bigcode/starcoder2data",
+        help="HF dataset path to the data",
+    )
+    parser.add_argument(
+        "--shard_size", type=int, default=1, help="Size, in GB, for each shard."
+    )
+    parser.add_argument(
+        "--data_dir", type=str, default=None, help="Optionally select a language"
```
blester125, line 60: I'm assuming this selects a subset of the dataset based on language? Can you add info on where to see the possible languages? Also, even if it is used as the "data_dir" later we should expose it to the user with a more self-descriptive name like "language".

blester125 (changes requested):
There are a few copy paste errors to clean up, but other than that it looks good to merge.

I'm also fine with merging it before the datasets go public

stackv2/sample.json
Muennighoff: Yeah it is how the dataset will look like once uploaded. Should I remove it?

stackv2/to_dolma.py
```diff
@@ -0,0 +1,64 @@
...
+        "--data_path",
+        default="bigcode/starcoder2data",
+        help="HF dataset path to the data",
+    )
+    parser.add_argument(
+        "--shard_size", type=int, default=1, help="Size, in GB, for each shard."
+    )
+    parser.add_argument(
+        "--language",
+        type=str,
```
blester125, line 59: Nit: `type=str` is the default, not needed.

blester125 (approved):

stackv2/sample.json
blester125: I think it fine either way, we just need to remember to remove it once the dataset is released.

stackv2/to_dolma.py
Muennighoff, line 59:
```suggestion
```

Muennighoff:
Should we preliminarily merge this @blester125 ? Will create a new PR then once the data is released to make the necessary changes.

PR #73, opened 2024-05-15
Library of Congress public domain books (loc_books)

storytracer:
This PR adds the code and documentation to download and export public domain books from the Library of Congress [Selected Digitized Books](https://www.loc.gov/collections/selected-digitized-books/) collection. The PR closes issue #74 .


loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+    stop_after_attempt,
+    wait_random_exponential,
+)
+from tqdm import tqdm
+
+from licensed_pile import logs
+from licensed_pile.licenses import PermissiveLicenses
+from licensed_pile.write import to_dolma
+
+data_path = Path(__file__).resolve().parent / "data"
```
blester125, line 28: Can we make this configurable based on cli arguments?

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+
+        logger.info(f"Found {len(text_file_urls)} books in metadata file")
+        logger.info(
+            f"{len(text_file_urls) - len(download_urls)} books already downloaded"
+        )
+        logger.info(f"Downloading {len(download_urls)} books")
+
+        self.progress_bar.total = len(download_urls)
+
+        with mp.Pool(10) as pool:
```
blester125, line 74: we should make this num threads configurable via cli

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+            results = pool.imap(functools.partial(self.download_book), download_urls)
```
blester125, line 75: Can you add a comment about how we don't actually use this `results` variable for more than updating the progress bar?

I thought there would be a bug based on consuming the iterable too early but I see that the real point is the download side effect

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+)
+from tqdm import tqdm
+
+from licensed_pile import logs
+from licensed_pile.licenses import PermissiveLicenses
+from licensed_pile.write import to_dolma
+
+data_path = Path(__file__).resolve().parent / "data"
+
+metadata_exports_path = data_path / "exports/metadata"
```
blester125, line 30: Globals like this should be all upper case, e.g., `METADATA_EXPORTS_PATH`

Would it be hard to move this into the class by setting them during the `__init__`? Then you could pass in `data_path` to make it easy to configure.

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+
+        with mp.Pool(10) as pool:
+            results = pool.imap(functools.partial(self.download_book), download_urls)
+            for result in results:
+                if result:
+                    self.progress_bar.set_description(f"{result['filename']}")
+                    self.progress_bar.update(1)
+                pass
+
+    def urls_to_download(self, text_file_urls):
```
blester125, line 82: Is there a reason we are pre-computing the files already downloaded? Would it be simpler to have a check in the download_book method and just log that a book was already downloaded and it is getting skipped?

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+            text_file_url
+            for text_file_url in text_file_urls
+            if Path(str(furl(text_file_url).path)).name not in existing_files
+        ]
+
+        return download_urls
+
+    def download_book(self, text_file_url):
+        text_file_furl = furl(text_file_url)
+        text_file_name = Path(str(text_file_furl.path)).name
```
blester125, line 96: Is the fact that we are using the `path` of the url mean we are going to create a bunch of directories? Would it be better to have a flat dir by sanitizing the path (replace `/` with `_` etc)?

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+        text_file_path = os.path.join(book_downloads_path, text_file_name)
```
blester125, line 97: We probably shouldn't mix `os.path` and `Path` in the same file. Can this be replaced with a `Path` operation?

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+        if not os.path.exists(text_file_path):
+            try:
+                for attempt in Retrying(
+                    stop=stop_after_attempt(3),
```
blester125, line 101: Even if we don't expose it to the user yet, should should have things like this `3` as a variable/parameter to this method instead of hardcoding it

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+        ]
+
+        return download_urls
+
+    def download_book(self, text_file_url):
+        text_file_furl = furl(text_file_url)
+        text_file_name = Path(str(text_file_furl.path)).name
+        text_file_path = os.path.join(book_downloads_path, text_file_name)
+        if not os.path.exists(text_file_path):
+            try:
```
blester125, line 99: Is there a simpler way to structure this? It seems like:
* The first try is the one that catches an error when you fail all your retries
* The inner try is for non-transient errors?

The loop makes things confusing, can you at least comment what all the bits are doing?

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+        filename = filepath.name
+        metadata = self.metadata.loc[self.metadata["text_file_name"] == filename]
+        if not metadata.empty:
+            with open(filepath) as f:
+                text = f.read()
+
+            dolma_data = {
+                "id": metadata["lccn"].values[0],
+                "text": text,
+                "source": "loc_books",
```
blester125, line 169: I think in other places we use a more verbose source name, like `Library of Congress Books`

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+                "added": datetime.datetime.utcnow().isoformat(),
+                "metadata": {
+                    "license": str(PermissiveLicenses.PD),
+                    "title": metadata["title"].values[0],
+                    "author": metadata["author"].values[0],
+                    "year": int(metadata["year"].values[0]),
+                    "language": metadata["language"].values[0],
+                    "item_url": f"https://www.loc.gov/item/{metadata['lccn'].values[0]}",
```
blester125, line 177: Probably just call this field `"url"`

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+                    "text_file_url": metadata["text_file_url"].values[0],
```
blester125, line 178: What's the difference between this and the item url? Can you document that?

loc_books/books.py
```diff
@@ -0,0 +1,213 @@
...
+    downloader = LocBooksDownloader(snapshot)
+    downloader.download()
+
+
+@main.command()
+@click.option("--snapshot", required=True, help="Snapshot name")
+@click.option("--dolma-shard-size", default=1, help="Shard file size in GB")
+@click.option(
+    "--dolma-filename",
+    required=True,
```
blester125, line 202: Probably doesn't need to be required

loc_books/metadata.py
```diff
@@ -0,0 +1,263 @@
...
+from dateutil.parser import parse
+from furl import furl
+from pyrate_limiter import Duration, Limiter, RequestRate
+from requests_ratelimiter import LimiterSession
+from titlecase import titlecase
+from tqdm import tqdm
+
+from licensed_pile import logs
+
+data_path = Path(__file__).resolve().parent / "data"
```
blester125, line 19: Same as above

loc_books/metadata.py
```diff
@@ -0,0 +1,263 @@
...
+            os.path.join(facet_folder, f"{str(1).zfill(filename_digits)}.json")
+        ) as f:
+            file_data = json.load(f)
+            total_pages = file_data["pagination"]["total"]
+
+        pages_to_download = self.check_existing_files(total_pages, facet_folder)
+        self.existing_pages_count += total_pages - len(pages_to_download)
+        self.progress_bar.total += len(pages_to_download)
+
+        with PoolExecutor(max_workers=10) as executor:
```
blester125, line 81: Is there a specific reason to use this over `multiprocessing.dummy.Pool` like you did before?

loc_books/metadata.py
```diff
@@ -0,0 +1,263 @@
...
+                    "start_year": start_year,
+                    "year_range": year_range,
+                    "count": filter["count"],
+                    "link": filter["on"],
+                }
+                date_facets.append(date_facet)
+
+        self.date_facets = date_facets
+
+    def check_existing_files(self, total_pages, output_folder):
```
blester125, line 111: Same question as above.

loc_books/metadata.py
```diff
@@ -0,0 +1,263 @@
...
+        pages_to_download = []
+        for page in range(1, total_pages + 1):
+            filename = os.path.join(
+                output_folder, f"{str(page).zfill(filename_digits)}.json"
+            )
+            if not os.path.exists(filename):
+                pages_to_download.append(page)
+        return pages_to_download
+
+    def download_page(self, facet_url, output_folder, page):
```
blester125, line 121: is there a reason we don't have retries like above?

blester125 (changes requested):
Thanks for the hard work! It looks really good and basically ready to merge!

Left a few comments, can you also create a script that calls all the steps in order?

PR #76, opened 2024-05-27
Hansard

baberabb:


hansard/source_code/ca_hansard.py
```diff
@@ -0,0 +1,106 @@
...
+                    pl.lit("[", dtype=pl.String),
+                    pl.col("speakerposition").str.to_titlecase(),
+                    pl.lit("]", dtype=pl.String),
+                    ignore_nulls=False,
+                )
+            ),
+        )
+        .lazy()
+        .group_by("speechdate")
+        .agg(
```
blester125, line 56: I assume this agg + concat_str is the reason we have the `ORDER BY` above? Is it guaranteed that order within a group is the same order that is within the full table? Or is there a way to have an order by that applies to the group like in SQL?

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+
+parser = argparse.ArgumentParser(description="Collect UK-Hansard into Dolma format")
+parser.add_argument(
+    "--base_folder",
+    default="hansard/uk_parlparse/scrapedxml",
+    help="Path to the directory of UK-Hansard XML files.",
+)
+parser.add_argument(
+    "--output_dir",
+    default="hansard/dolma_outputs",
```
blester125, line 22: this should probably default to something like `data/hansard/v0`

hansard/ca_to_dolma.py
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Collect Canadian Hansard into Dolma format."
+    )
+    parser.add_argument(
+        "--output_folder",
+        default="hansard/ca_hansard",
```
blester125, line 37: This should probably default to something like `data/hansard/v0`

hansard/ca_to_dolma.py
```diff
@@ -0,0 +1,64 @@
...
+        help="Output format for parquet files",
+    )
+    parser.add_argument(
+        "--file-name",
+        default="cahansard.jsonl.gz",
```
blester125, line 42: probably a default like `ca_hansard.json.gz` is more understandable

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+                yield from process_files_in_folder(subfolder, source)
+
+
+def main(args):
+    base_folder = Path(args.base_folder)
+    to_dolma(
+        examples=process_folder(base_folder, count=args.count),
+        path=args.output_dir,
+        shard_size=args.shard_size,
+        filename="ukhansard.jsonl.gz",
```
blester125, line 156: This should probably be an argument like the CA version, and probably be `uk_hansard.jsonl.gz` to be a bit clearer.

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+    subfolders = [f for f in folder_path.iterdir() if f.is_dir()]
+    return subfolders
+
+
+def parse_hansard_xml_file(root: ET._Element) -> str:
+    parsed_text = []
+    for element in root.iter():
+        if element.tag == "major-heading":
+            text = ET.tostring(element, method="text", encoding="unicode").strip()
+            parsed_text.append(f"{text}")
```
blester125, line 79: Why does this need to be an fstring?

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+        elif element.tag in ["oral-heading", "minor-heading"]:
```
blester125, line 80: This seems to be processed the same way as the `major-heading`, can these branches be combined?

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+            text = ET.tostring(element, method="text", encoding="unicode").strip()
+            parsed_text.append(f"{text}")
+        elif element.tag in [
```
blester125, line 83: Read-only sequences like this should be a tuple, maybe even a frozenset given we are doing an `in` query

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+            "speech",
+            "ques",
+            "reply",
+            "question",
+        ]:
+            speaker = (
```
blester125, line 89:
```python
speaker = f"{s}: " if (s := element.attrib.get("speakername", "")) else ""
```

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+                element.attrib.get("speakername", "") + ": "
+                if element.attrib.get("speakername", "")
+                else ""
+            )
+            speech_text = (
+                ET.tostring(element, method="text", encoding="unicode")
+                .strip()
+                .replace("    ", "")
```
blester125, line 97: This would probably be more robust if it was a regex replacing more than one whitespace with a single whitespace.

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+            )
+            parsed_text.append(f"{speaker + speech_text}")
```
blester125, line 99: You should be using the f-string to do your formatting, currently, speaker ends with a space (`": "`) and then you are adding it here, instead remove the trailing space from speaker and do something like
```python
parsed_text.append(f"{speaker} {speech_text}".strip())
```

Otherwise the fstring is pointless here.

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return "\n\n".join(parsed_text).replace(" ", "")
```
blester125, line 101: Isn't this `.replace` going to remove all whitespace? again I think a regex for more than one whitespace would be better.

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def process_files_in_folder(
+    folder_path: Path, source: str, count: bool = False
+) -> dict:
+    global COUNT
+    language = "en" if not source == "senedd-cy" else "cy"
+    date_match = re.compile(r"\d{4}-\d{2}-\d{2}")
```
blester125, line 109: Regex's get compiled and cached under the hood so this probably isn't needed

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+    for file in folder_path.iterdir():
+        if file.suffix == ".xml" and file.stem != "tmp":
+            root = ET.parse(file, parser=PARSER).getroot()
+            parsed_text = parse_hansard_xml_file(root)
+            if parsed_text:
+                if count:
+                    COUNT += len(WHITESPACE.split(parsed_text))
+                date_ = date_match.search(file.stem)
```
blester125, line 117: This is one of the best places for the walrus operator:

```python
date = d.group() if (d := date_match.search(file.stem)) else None
```

Also, I think `None` is a better default than a year far in the future

hansard/uk_to_dolma.py
```diff
@@ -0,0 +1,164 @@
...
+                        "license": str(PermissiveLicenses.OPL),
+                        "language": language,
+                        "year": date.split("-")[0],
+                    },
+                }
+
+
+def process_folder(folder_path: Path, count: bool = False) -> Iterator[dict]:
+    for subfolder in get_subfolders(folder_path):
+        if subfolder.name in FILE_NAMES:
```
blester125, line 138: The `FILE_NAMES` datastructure is kinda confusing, I think we don't really need the `source` key nesting, it should just be filename -> source name.

Instead of special casing `senedd` _in_ the loop, we should just `continue` when `subfolder.name == senedd` and then run the nested_subfolder loop later as a special case.

blester125 (changes requested):
Listed out some changes for readability and organization. I think we should also remove all the `COUNT` global variable stuff. I can whip up a counting tool with dolma pretty quick

hansard/ca_to_dolma.py
```diff
@@ -0,0 +1,64 @@
...
+from licensed_pile.licenses import PermissiveLicenses
+from licensed_pile.write import to_dolma
+
+WHITESPACE = regex.compile(r"\w+|[^\w\s]+")
+COUNT = 0
+
+
+def format_dolma(row: dict) -> dict:
+    global COUNT
+    COUNT += len(WHITESPACE.split(row["text"]))
```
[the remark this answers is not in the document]
baberabb: whoops! this was temporary. meant to delete!

hansard/source_code/ca_hansard.py
baberabb: yeah. Within groups the order of the rows is always preserved

baberabb:
@blester125 incorporated all your comments. Thanks!

baberabb:
@blester125 This should be good to merge as well!

Issue #21, opened 2023-11-07
Legal Documents


blester125:
Domain: Legal

- [ ] Pile of Law
- [ ] Case Law Access Project

- [ ] US Congressional Documents 
Digitized records of congress proceedings. See here: https://www.govinfo.gov/app/collection/cdoc/118/sdoc/all
Some of the data is text and some is just PDF, from a quick look it seems like there are a decent number of tables in the PDFs (which generally don't have text versions available).

StellaAthena:
I have all pre-2020 Congressional proceedings, but when we read over them we quickly decided that the median document was far to racist to be included in a training dataset.

conceptofmind:
> > I have all pre-2020 Congressional proceedings, but when we read over them we quickly decided that the median document was far to racist to be included in a training dataset.
> 
> @StellaAthena do we want to intentionally exclude toxic content, or to include it but with toxicity/quality scores attached? Based on prior work I wonder if it's better to let users decide, especially since there are beneficial cases to pretrain on real world toxic content (e.g. training realistic toxicity detectors).

We could use perspectives api to list the scores for each sentence.

conceptofmind:
Also should keep Cushman in the loop for this? Talk to Peter as well.

craffel:
Regarding toxicity annotations, we are not going to do that as part of the per-source preprocessing, it will be done globally to all sources.

conceptofmind:
> Regarding toxicity annotations, we are not going to do that as part of the per-source preprocessing, it will be done globally to all sources.

Got you. Sounds good.

conceptofmind:
I have access to all of case law and some scripts to get the data. Can CC John Nay

wildphoton:
@conceptofmind Hi Enrico, if you have not started to work on Pile of Law, I will take it.

StellaAthena:
@wildphoton Note that the Pile of Law is an amalgamation of different data sources. We should process those from their source rather than just import the Pile of Law.

Looking through the paper, it seems that about two thirds of the data comes from:
1. CourtListener Opinions, CourtListener Docket Entries and Court Filings
2. U.S. Board of Veterans’ Appeals Decisions
3. Atticus Contracts
4. Edgar Contract
5. U.S. State Codes

wildphoton:
> @wildphoton Note that the Pile of Law is an amalgamation of different data sources. We should process those from their source rather than just import the Pile of Law.
> 
[...]

PR #83, opened 2024-06-04
Add correct text ordering and column merging to CL

conceptofmind:
The columns in CL need to be coalesced in the order:
```
"html_with_citations"
"html_columbia"
"html_lawbox"
"xml_harvard"
"html_anon_2020"
"html"
"plain_text"
```
The `plain_text` only contains a smaller low-quality subset of the total data. Using only the `plain_text` column misses much of the information that is available in the other columns. Not all rows contain information. To get the largest quantity of proper CL data possible without additionally scraping the `.txt` files it is required to merge the columns in the order above.

Harvard in cold cases and CAP chose to use regex to do initial text extraction. This can be replaced with something like Trafilatura in one line.

I tried to keep this similar to what zhenlinx had. I did not remove their script.

The script: 
- Reads the decompressed CL data dump 
- Adds metadata and source information
- Drops unnecessary columns based on Harvard CAP input
- Combines columns in the correct order based on CL documentation
- Drops repeated XML/HTML columns
- Extracts HTML/XML following Harvard CAP
- Merges extracted text and plain text properly
- Handles null values based on text row
- Writes out to dolma format

I iterated through the `jsonl.gz` file to make sure it looked ~normal.

Additional filtering and boilerplate removal will need to be added in another PR.

This leaves you with the "correct" initial CL data following guidelines by Harvard CAP and CL/Mike Lissner.

@blester125


conceptofmind:
A side note. If I am to add Spark processing I will need to make additional adjustments to other parts of the scripts.

craffel:
Thanks for this. I don't see any reason to have both versions of the processing pipelines, can you merge this into the existing one?

courtlistener/process_cl.py
blester125, line 1: This should be a modification of the csv_to_dolma.py instead of a new file

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    to_dolma(example, args.output_dir, output_file_base_name, args.shard_size)
+    logger.info(f"Saved {args.input_file} as dolma shared files at {args.output_dir}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Convert csv data to dolma.")
+    parser.add_argument(
+        "--output_dir",
+        default=f"data/courtlistener/v0",
```
blester125, line 118: This should be `"data/courtlistener/v0/documents"`

blester125, line 118: This doesn't need to be an f-string

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+        help="Where the dolma formatted data goes.",
+    )
+    parser.add_argument(
+        "--shard_size",
+        default=1000,
+        help="The number of documents to store in each shard.",
+    )
+    parser.add_argument(
+        "--input_file",
+        default="./data/courtlistener/raw/opinions-2022-08-02.csv",
```
blester125, line 128: We shouldn't have a default for this, especially as the date doesn't match the latest date. Lets use `required=True`  and remove the default.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    # drop plain text column and text null values
+    df = df.drop(columns=["plain_text"]).dropna(subset=["text"])
+
+    # return a dictionary for each row - dolma format
+    return df.to_dict(orient="records")
+
+
+def main(args):
+    example = process_court_listener(args.input_file)
+    output_file_base_name = os.path.basename(args.input_file).replace(
```
blester125, line 107: Now that we don't have multiple dated files we should have the base filename as an cli argument with a default like `courtlistener.jsonl.gz` instead of basing it on the input file

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    df["text"] = df["text"].apply(lambda x: re.sub(r"<.+?>", "", x))
+
+    # combine merge plain text and extracted text
+    df["text"] = df["text"].combine_first(df["plain_text"])
+
+    # drop plain text column and text null values
+    df = df.drop(columns=["plain_text"]).dropna(subset=["text"])
+
+    # return a dictionary for each row - dolma format
+    return df.to_dict(orient="records")
```
blester125, line 102: This doesn't result in the dolma formatted data. See https://github.com/allenai/dolma/blob/main/docs/data-format.md#dolma-document-format

Issues include:

* It has a `"download_url"` key top-level. This should be removed or moved into metadata as `"url"`
* It has `"date_created"` instead of `"created"`
* It is missing an `"added"` field
* It has the value of `"metadata"` as the license string. It should be a dict with a `"license"` key instead.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    # add metadata column
+    df["metadata"] = str(PermissiveLicenses.PD)
+
+    # add source column
+    df["source"] = SOURCE_NAME
+
+    """
+    Remove columns:
+    date_modified
+    author_str
```
blester125, line 33: We should include the `author_str` as an `"author"` field in the `"metadata"` dict.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+            "page_count",
+            "local_path",
+            "extracted_by_ocr",
+            "author_id",
+            "cluster_id",
+        ]
+    )
+
+    """
+    Merge columns based on Court Listener documentation:
```
blester125, line 61: We should re-write this comment. "Merge" isn't the name for this operation. Something like:

```
For each row, select the first non-null value when iterating over columns in this order: ...
This follows the procedure in the Court Listener documentation.
```

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+            "sha1",
+            "page_count",
+            "local_path",
+            "extracted_by_ocr",
+            "author_id",
+            "cluster_id",
+        ]
+    )
+
+    """
```
blester125, line 60: Prefer multiple single line comments over multiline strings as comments (syntax highlighting doesn't handle the string as a comment)

https://peps.python.org/pep-0008/#block-comments

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    type
+    sha1
+    page_count
+    local_path
+    extracted_by_ocr
+    author_id
+    cluster_id
+    """
+    df = df.drop(
+        columns=[
```
blester125, line 45: Not something to be addressed, but I hate that pandas would do something totally different if this was a tuple instead of a list (and it would be better as a tuple as it is read only lol)

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+def process_court_listener(file_path):
+    df = pd.read_csv(file_path)
+
+    # add metadata column
+    df["metadata"] = str(PermissiveLicenses.PD)
+
+    # add source column
+    df["source"] = SOURCE_NAME
+
+    """
```
blester125, line 30: This comment doesn't add anything, the list of columns to remove is in the code below

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def process_court_listener(file_path):
+    df = pd.read_csv(file_path)
+
+    # add metadata column
+    df["metadata"] = str(PermissiveLicenses.PD)
+
+    # add source column
+    df["source"] = SOURCE_NAME
```
blester125, line 28: Lets add a `source_name` argument (which we can default to `SOURCE_NAME`) and set that instead of using a global.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+csv.field_size_limit(sys.maxsize)
+
+logger = configure_logging("court-listener-opinion")
+
+
+def process_court_listener(file_path):
+    df = pd.read_csv(file_path)
+
+    # add metadata column
+    df["metadata"] = str(PermissiveLicenses.PD)
```
blester125, line 25: The metadata value should be a dict in the dolma format, not a string

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+import pandas as pd
+
+from licensed_pile.licenses import PermissiveLicenses
+from licensed_pile.logs import configure_logging
+from licensed_pile.write import to_dolma
+
+SOURCE_NAME = "CourtListenerOpinion"
+
+csv.field_size_limit(sys.maxsize)
```
blester125, line 16: Is this needed now that we aren't using the csv stdlib?

If it isn't there are several unused imports to remove.

courtlistener/process_cl.py
blester125, line 1: We should add a simple docstring to the module

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
+import argparse
+import csv
+import logging
```
blester125, line 3: Unneeded import

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+from licensed_pile.licenses import PermissiveLicenses
+from licensed_pile.logs import configure_logging
+from licensed_pile.write import to_dolma
+
+SOURCE_NAME = "CourtListenerOpinion"
+
+csv.field_size_limit(sys.maxsize)
+
+logger = configure_logging("court-listener-opinion")
```
blester125, line 18: prefer moving the configure_logging call under `if __name__ == "__main__":` and then have a `logger = logs.get_logger(...)` in the `process_court_listener` function.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def process_court_listener(file_path):
```
blester125, line 21: Missing Docstring

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    df = df.drop(
+        columns=[
+            "html",
+            "html_anon_2020",
+            "html_lawbox",
+            "html_columbia",
+            "xml_harvard",
+            "html_with_citations",
+        ]
+    ).dropna(subset=["text"])
```
blester125, line 89: Drop NA is a very different operation than dropping columns. We should separate them into two different lines so the dropna doesn't get lost in a refactor.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # extract text from html and xml following Harvard CAP
+    # They used r"<.+?>", ""
+    df["text"] = df["text"].apply(lambda x: re.sub(r"<.+?>", "", x))
```
blester125, line 93: This can get vectorized via built in pandas instead of using apply, i.e.,:

```python
df["text"] = df["text"].str.replace(r"<.+?>", "", regex=True)
```

conceptofmind:
> Thanks for this. I don't see any reason to have both versions of the processing pipelines, can you merge this into the existing one?

Sounds good. I will merge it into the existing one.

courtlistener/process_cl.py
blester125, line 93: Do we have any auditing tools for the output? It can be ok if you have a strict well-defined subset of things you are looking for, but generally you shouldn't parse HTML/XML with a regex https://stackoverflow.com/a/1732454

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # combine merge plain text and extracted text
```
blester125, line 95: Again, merge is a bad name for this operation.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+    df["text"] = df["text"].combine_first(df["plain_text"])
```
blester125, line 96: This line doesn't do anything, there is already a `.dropna(subset=["text"])` call so there won't be any line where the `text` column is `None` and the `plain_text` column is non-null.

Additionally, `.combine_first` doesn't consider an empty string as null so even if there was a `text` column that was totally removed based on the regex above, it wouldn't get replaced by the `plain_text` column.

The `plain_text` column should be part of the initial `.combine_first` chain or the `.dropna` above should be removed and rows with an empty string for `text` column should be converted to a `None` so the `plain_text` column will actually replace the `text` column

blester125 (changes requested):
I listed out several changes that should be made, currently the output format is not correct.

I can't see needing something like SPARK for this dataset. If there are OoM or slowness issues due to pandas reading the whole CSV at once it should be easy to update to 1) Read the CSV line-by-line via a generator that uses the csv from stdlib 2)  parallelized the conversion from csv record to dolma record with `multiprocessing.Pool.imap` 3) pass the resulting iterator to the `to_dolma` function.

That should avoid reading the whole file at once and leverage the embarrassingly parallel nature of the task.

conceptofmind:
> I listed out several changes that should be made, currently the output format is not correct.
> 
> I can't see needing something like SPARK for this dataset. If there are OoM or slowness issues due to pandas reading the whole CSV at once it should be easy to update to 1) Read the CSV line-by-line via a generator that uses the csv from stdlib 2) parallelized the conversion from csv record to dolma record with `multiprocessing.Pool.imap` 3) pass the resulting iterator to the `to_dolma` function.
> 
> That should avoid reading the whole file at once and leverage the embarrassingly parallel nature of the task.

I tried to keep it as close to the original as possible. There are numerous different changes I would have personally made. I can integrate those instead as well.

blester125:
Also, if you are going to keep working on the same branch after a merge (which you should avoid), you should prefer rebasing onto main instead of merging it into your branch (which happened here https://github.com/r-three/common-pile/pull/83/commits/9df6741fb085c6d93a3acc604c20eb1397f8679b)

conceptofmind:
I will resolve each of the comments. However, the text extractor should be decided upon before I do so. This is a minimal reproduction of what was done in Spark.

conceptofmind:
> Also, if you are going to keep working on the same branch after a merge (which you should avoid), you should prefer rebasing onto main instead of merging it into your branch (which happened here [9df6741](https://github.com/r-three/common-pile/commit/9df6741fb085c6d93a3acc604c20eb1397f8679b))

That was a "fat finger" haha. Noticed after it and hadn't squashed.

conceptofmind:
@blester125 Do you have opinion on trafilatura?

courtlistener/process_cl.py
conceptofmind: Yes will merge.

courtlistener/process_cl.py
conceptofmind: I do not think much of this is needed.

courtlistener/process_cl.py
conceptofmind: Will do.

courtlistener/process_cl.py
conceptofmind: This is an artifact of using a different extractor.

conceptofmind: Trafilatura will over-filter and cause null if not handled explicitly.

conceptofmind: This is not the case with standard regex.

courtlistener/process_cl.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # drop plain text column and text null values
+    df = df.drop(columns=["plain_text"]).dropna(subset=["text"])
```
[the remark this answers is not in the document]
conceptofmind: Can split and add comments.

courtlistener/process_cl.py
conceptofmind: This is dependent on the extractor we choose.

courtlistener/process_cl.py
conceptofmind: I can list it as coalesce.

courtlistener/process_cl.py
conceptofmind: Got it.

conceptofmind: Got it.

courtlistener/process_cl.py
conceptofmind: Will do.

courtlistener/process_cl.py
conceptofmind: Will do.

courtlistener/process_cl.py
conceptofmind: Will do and add comments.

courtlistener/process_cl.py
conceptofmind: Coalesce is what is used in spark.

conceptofmind: I have no opinion on this. We can add that comment.

courtlistener/process_cl.py
conceptofmind: Will do.

courtlistener/process_cl.py
conceptofmind: I will create a metadata dict for the recommended changes.

courtlistener/process_cl.py
conceptofmind: Will change.

courtlistener/process_cl.py
conceptofmind: WIll do.

courtlistener/process_cl.py
conceptofmind: Will create a metadata dict as noted above.

courtlistener/process_cl.py
blester125: If Trafilatura is so overzelous that it turns real data into null that sounds like a red flag. Also if the `plain_text` column is good enough for those cases is there a reason it isn't good enough to always use it?

conceptofmind:
@blester125 

Given the comments on parallelization. 

The csv can be iterated through every 20k rows following something like the below:

```
csv_file = f'{opinions_string}.csv'

chunk_size = 20000 

# Read the CSV file in chunks
for i, chunk in enumerate(pd.read_csv(csv_file, chunksize=chunk_size)):
    # processing code
```

I am not sure if this is what you had in mind.

Harvard CAP handled loading the entire file initially and then repartitioning.

There are a few ways to handle the csv and I am not opinionated. 

We can also preprocess the large csv into smaller files as a separate step.

courtlistener/process_cl.py
conceptofmind: I agree that it is a red flag. I am writing my own explicit extractor for this dataset but that will not be out until later.

Trafilatura ranks as the highest-quality extractor given a general task.

The extractor should be run on every column which is not plain text. Plain text is considered the lowest quality data in terms of coalescing. It should only be used if the other columns do not exist and can not be extracted.

courtlistener/process_cl.py
conceptofmind: I will add a docstring following:

```
Reads the decompressed CL data dump
Adds metadata and source information
Drops unnecessary columns based on Harvard CAP input
Combines columns in the correct order based on CL documentation
Drops repeated XML/HTML columns
Extracts HTML/XML following Harvard CAP
Merges extracted text and plain text properly
Handles null values based on text row
Writes out to dolma format
```

blester125:
The most straightforward way of parallelism would probably be something like this:

```python
data = read_csv_generator(csv_file)
with mp.Pool(args.processors) as pool:
  dolma = pool.imap(convert_record_to_dolma, data)
  to_dolma(dolma, ...)
```

You could either have the read_csv_generator output individual rows (and then pandas probably isn't the best fit) or you could have it output chunks like you mentioned (as long as passing the pandas dataframe to the new process isn't extremely expensive) and would just need to add something like `dolma = itertools.chain(*dolma)` before the `to_dolma` call.

Assuming there isn't too much contention wrt to the disk, I would think this could saturate a multicore machine. TBH I don't think this dataset is big enough that we need to be 100 percent efficient

conceptofmind:
> The most straightforward way of parallelism would probably be something like this:
> 
> ```python
> data = read_csv_generator(csv_file)
> with mp.Pool(args.processors) as pool:
>   dolma = pool.imap(convert_record_to_dolma, data)
>   to_dolma(dolma, ...)
> ```
> 
> You could either have the read_csv_generator output individual rows (and then pandas probably isn't the best fit) or you could have it output chunks like you mentioned (as long as passing the pandas dataframe to the new process isn't extremely expensive) and would just need to add something like `dolma = itertools.chain(*dolma)` before the `to_dolma` call.
> 
> Assuming there isn't too much contention wrt to the disk, I would think this could saturate a multicore machine. TBH I don't think this dataset is big enough that we need to be 100 percent efficient

One dump is around 240 GB csv decompressed. Many mid-sized machines could load it and process in one go.

wildphoton:
Use pandas, even read in chunks, won't avoid having all rows in memory IMO. If we expect people will run our code, shall we make the code more memory friendly instead of speed efficient? Previously, I read and process the file into dolma line by line on the fly and you can do what ever you want in the `process_row()`. Although this is not parallelized, it is not too slow to iterate the whole file. How do you think @conceptofmind @blester125 

```
with open(file_path, "r") as csvfile:
    reader = csv.DictReader(csvfile)

    # Yield a dictionary for each row
    for row in reader:
        dolma_dict = process_row(row)
        yield row
```

conceptofmind:
> Use pandas, even read in chunks, won't avoid having all rows in memory IMO. If we expect people will run our code, shall we make the code more memory friendly instead of speed efficient? Previously, I read and process the file into dolma line by line on the fly and you can do what ever you want in the `process_row()`. Although this is not parallelized, it is not too slow to iterate the whole file. How do you think @conceptofmind @blester125
> 
> ```
> with open(file_path, "r") as csvfile:
>     reader = csv.DictReader(csvfile)
> 
>     # Yield a dictionary for each row
>     for row in reader:
>         dolma_dict = process_row(row)
>         yield row
> ```

The above `pandas read in chunks` runs on my local desktop which does not even have close to enough RAM to load the entire 250GB into memory. I would not think that the pandas read csv would handle mem mapping that well to fit such a large dataset into memory but I could be wrong.

The read in chunks should function similarly to a iterator.

I also have no strong opinion on this.

courtlistener/process_cl.py
conceptofmind:
```python
to_dolma(example, args.output_dir, args.output_file_base_name, args.shard_size)

...

parser.add_argument(
  "--output_file_base_name",
  default="courtlistener.jsonl.gz",
  help="The base name of the output file.",
)
```

courtlistener/process_cl.py
conceptofmind:
```python
parser.add_argument(
    "--input_file",
    required=True,
    help="The path to the csv file to convert.",
)
```

courtlistener/process_cl.py
conceptofmind:
```python
    parser.add_argument(
        "--output_dir",
        default="data/courtlistener/v0/documents",
        help="Where the dolma formatted data goes.",
    )
```

courtlistener/process_cl.py
conceptofmind:
```python
    # For each row, select the first non-null value when iterating over columns in this order:
    # html_with_citations
    # html_columbia
    # html_lawbox
    # xml_harvard
    # html_anon_2020
    # html
    # plain_text
    # This follows the procedure in the Court Listener documentation.

    df["text"] = (
        df["html_with_citations"]
        .combine_first(df["html_columbia"])
        .combine_first(df["html_lawbox"])
        .combine_first(df["xml_harvard"])
        .combine_first(df["html_anon_2020"])
        .combine_first(df["HTML"])
    )
```

courtlistener/process_cl.py
conceptofmind:
```python
    # For each row, select the first non-null value when iterating over columns in this order:
    # html_with_citations
    # html_columbia
    # html_lawbox
    # xml_harvard
    # html_anon_2020
    # html
    # plain_text
    # This follows the procedure in the Court Listener documentation.

    df["text"] = (
        df["html_with_citations"]
        .combine_first(df["html_columbia"])
        .combine_first(df["html_lawbox"])
        .combine_first(df["xml_harvard"])
        .combine_first(df["html_anon_2020"])
        .combine_first(df["html"])
    )
```

courtlistener/process_cl.py
conceptofmind:
```python
    parser.add_argument(
        "--output_dir",
        default="data/courtlistener/v0/documents",
        help="Where the dolma formatted data goes.",
    )
```

courtlistener/process_cl.py
conceptofmind:
```python
def process_court_listener(file_path):
    ...
    df["source"] = args.source_name
```
```python
    parser.add_argument(
        "--source_name",
        default="SOURCE_NAME",
        help="The name of the source.",
    )
```

courtlistener/process_cl.py
conceptofmind: Deleted this and removed imports.

courtlistener/process_cl.py
conceptofmind: Removed import

courtlistener/process_cl.py
conceptofmind: I think this is what you meant:
```python
def process_court_listener(file_path):

    logger = logging.getLogger("court-listener-opinion")
    ...
```
```python
if __name__ == "__main__":

    logger = configure_logging("court-listener-opinion")
    ...
```

courtlistener/process_cl.py
conceptofmind: Maybe something like this:
```
    # Combine DataFrame objects by filling null values 
    # with non-null values from the other selected DataFrames.
```

courtlistener/process_cl.py
conceptofmind: Removed comment

```python
def process_court_listener(file_path):
    ...
    df["source"] = args.source_name
```

courtlistener/process_cl.py
conceptofmind: How about something like:
```python
def process_court_listener(file_path):
    """
    This function performs several operations on a given file:

    1. Reads the decompressed Court Listener (CL) data dump from the provided file path.
    2. Adds metadata and source information to the data.
    3. Drops unnecessary columns based on Harvard CAP input.
    4. Selects the first non-null value when iterating over columns based on CL documentation.
    5. Drops repeated XML/HTML columns.
    6. Extracts HTML/XML following Harvard CAP guidelines.
    7. Selects the first non-null value when iterating over extracted text and plain text columns.
    8. Handles null values based on text row.
    9. Writes out the processed data to dolma format.

    Parameters:
    file_path (str): The path to the file to be processed.
    """
```

courtlistener/process_cl.py
conceptofmind: How about something like this:
```python
    # drop the columns used in combine_first to avoid redundancy
    df = df.drop(
        columns=[
            "html",
            "html_anon_2020",
            "html_lawbox",
            "html_columbia",
            "xml_harvard",
            "html_with_citations",
        ]
    )

    # drop null values in the text column
    df = df.dropna(subset=["text"])
```

courtlistener/process_cl.py
conceptofmind:
```python
    # Combine DataFrame objects by filling null values 
    # with non-null values from the other selected DataFrames.
    df["text"] = df["text"].combine_first(df["plain_text"])
```

courtlistener/process_cl.py
conceptofmind: This is what it would be with trafilatura:
```python
    # extract text from html and xml using trafilatura
    df["text"] = df["text"].apply(lambda x: extract(x))
```

courtlistener/process_cl.py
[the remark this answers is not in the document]
conceptofmind:
```python
    # drop the plain text column
    df = df.drop(columns=["plain_text"])
    
    # drop null values in the text column
    df = df.dropna(subset=["text"])
```

courtlistener/process_cl.py
conceptofmind: Let me know if this is what you were looking for:
```python
    df = pd.read_csv(file_path)

    df["source"] = args.source_name
    df["added"] = datetime.utcnow().isoformat()

    # rename the column date_created -> created
    df = df.rename(columns={"date_created": "created"})

    # add the metadata column including the author_str, license, and url
    df["metadata"] = df.apply(
        lambda x: {
            "license": str(PermissiveLicenses.PD),
            "url": x["download_url"],
            "author": x["author_str"],
        },
        axis=1,
    )
```

courtlistener/process_cl.py
conceptofmind: Added this:
```python
"author": x["author_str"],
```

courtlistener/process_cl.py
conceptofmind: Added to metadata dict:
```python
            "license": str(PermissiveLicenses.PD),
```

courtlistener/process_cl.py
conceptofmind: LOL

conceptofmind:
@blester125 Responded to each of the comments.

An updated example script for loading the entire file would look like:
```python
import argparse
import logging
import re
from datetime import datetime
import pandas as pd

from licensed_pile.licenses import PermissiveLicenses
from licensed_pile.logs import configure_logging
from licensed_pile.write import to_dolma


def process_court_listener(file_path):
    """
    This function performs several operations on a given file:

    1. Reads the decompressed Court Listener (CL) data dump from the provided file path.
    2. Adds metadata and source information to the data.
    3. Drops unnecessary columns based on Harvard CAP input.
    4. Selects the first non-null value when iterating over columns based on CL documentation.
    5. Drops repeated XML/HTML columns.
    6. Extracts HTML/XML following Harvard CAP guidelines.
    7. Selects the first non-null value when iterating over extracted text and plain text columns.
    8. Handles null values based on text row.
    9. Writes out the processed data to dolma format.

    Parameters:
    file_path (str): The path to the file to be processed.
    """

    logger = logging.getLogger("court-listener-opinion")

    df = pd.read_csv(file_path)

    df["source"] = args.source_name
    df["added"] = datetime.utcnow().isoformat()

    # rename the column date_created -> created
    df = df.rename(columns={"date_created": "created"})

    # add the metadata column including the author_str, license, and url
    df["metadata"] = df.apply(
        lambda x: {
            "license": str(PermissiveLicenses.PD),
            "url": x["download_url"],
            "author": x["author_str"],
        },
        axis=1,
    )

    df = df.drop(
        columns=[
            "date_modified",
            "author_str",
            "per_curiam",
            "joined_by_str",
            "type",
            "sha1",
            "page_count",
            "local_path",
            "extracted_by_ocr",
            "author_id",
            "cluster_id",
            "download_url",
        ]
    )

    # For each row, select the first non-null value when iterating over columns in this order:
    # html_with_citations
    # html_columbia
    # html_lawbox
    # xml_harvard
    # html_anon_2020
    # html
    # plain_text
    # This follows the procedure in the Court Listener documentation.

    df["text"] = (
        df["html_with_citations"]
        .combine_first(df["html_columbia"])
        .combine_first(df["html_lawbox"])
        .combine_first(df["xml_harvard"])
        .combine_first(df["html_anon_2020"])
        .combine_first(df["html"])
    )

    # drop the columns from combine_first to avoid redundancy
    df = df.drop(
        columns=[
            "html",
            "html_anon_2020",
            "html_lawbox",
            "html_columbia",
            "xml_harvard",
            "html_with_citations",
        ]
    )

    # drop null values in the text column
    df = df.dropna(subset=["text"])

    # extract text from html and xml following Harvard CAP
    # They used r"<.+?>", ""
    df["text"] = df["text"].apply(lambda x: re.sub(r"<.+?>", "", x))

    # Combine DataFrame objects by filling null values
    # with non-null values from the other selected DataFrames.
    df["text"] = df["text"].combine_first(df["plain_text"])

    # drop the plain text column
    df = df.drop(columns=["plain_text"])

    # drop null values in the text column
    df = df.dropna(subset=["text"])

    # return a dictionary for each row - dolma format
    return df.to_dict(orient="records")


def main(args):
    example = process_court_listener(args.input_file)
    to_dolma(example, args.output_dir, args.output_file_base_name, args.shard_size)
    logger.info(f"Saved {args.input_file} as dolma shared files at {args.output_dir}")


if __name__ == "__main__":
    logger = configure_logging("court-listener-opinion")

    parser = argparse.ArgumentParser(description="Convert csv data to dolma.")
    parser.add_argument(
        "--output_dir",
        default="data/courtlistener/v0/documents",
        help="Where the dolma formatted data goes.",
    )
    parser.add_argument(
        "--shard_size",
        default=1000,
        help="The number of documents to store in each shard.",
    )
    parser.add_argument(
        "--input_file",
        required=True,
        help="The path to the csv file to convert.",
    )
    parser.add_argument(
        "--output_file_base_name",
        default="courtlistener.jsonl.gz",
        help="The base name of the output file.",
    )
    parser.add_argument(
        "--source_name",
        default="SOURCE_NAME",
        help="The name of the source.",
    )
    args = parser.parse_args()
    main(args)
```
Text extraction comment related to trafilatura needs to be decided. If not using trafilatura can remove the other `combine_first`.

conceptofmind:
If you want to iterate then something like this should work:
```
    for chunk in pd.read_csv(file_path, chunksize=args.shard_size):
        df = chunk
        ...
        yield df.to_dict(orient="records")
```

blester125:
Yeah, I think reading in chunks and passing the chunks to a multiprocessing pool would be the way to go

conceptofmind:
> Yeah, I think reading in chunks and passing the chunks to a multiprocessing pool would be the way to go

Ok I will add that to the main function.

conceptofmind:
I tested something like this given your comment:
```python
def iterate_over_chunks(file_path):
    for chunk in pd.read_csv(file_path, chunksize=args.shard_size):
        yield chunk

def main(args):
    data = iterate_over_chunks(args.input_file)
    with mp.Pool(args.processors) as pool:
        dolma_chunks = pool.imap(process_court_listener, data)
        dolma = itertools.chain(*dolma_chunks)
        to_dolma(dolma, args.output_dir, args.output_file_base_name, args.shard_size
```
The speedup is not massively significant. Limitations of pandas may be shown here.

As noted by Zhenlin Xu. Single-process is still reasonable and can be done quickly as well. Even with just iterating alone.

conceptofmind:
> I see, I think this version makes sense, even if the speedup isn't the best. Lets get all the changes into the PR and get it merged!

Will do

wildphoton:
I have tried to use BeautifulSoup for html/xml parsing and it seems working well. I will share the processed data soon. @conceptofmind @blester125

conceptofmind:
> I have tried to use BeautifulSoup for html/xml parsing and it seems working well. I will share the processed data soon. @conceptofmind @blester125

Hi,

Pre-processing steps need to be applied to the HTML/XML before text extraction. CourtListener contains malformed HTML/XML that requires upstreamed content from CAP. 

BS4 uses LXML as a backend which is typically worse quality than Trafilatura. I have already processed all of the data and uploaded the newest dump. I am waiting on HF to remove my rate limits so that I can finish it.

wildphoton:
> Hi,
> 
> Pre-processing steps need to be applied to the HTML/XML before text extraction. CourtListener contains malformed HTML/XML that requires upstreamed content from CAP.
> BS4 uses its own LXML backend extraction which is typically worse quality than Trafilatura extraction when precision is set True. I have already processed all of the data and uploaded the newest dump. I am waiting on HF to remove my rate limits so that I can finish it.


What kind of preprocessing steps are needed? What upstream content from CAP does it require? The current PR does not show these. I just applied BS4 text extraction on the raw data and it works great. If Trafilatura works better, could you include it and all the preprocessing you mentioned in the CR? A stand-alone dataset won't be very useful. Thanks.

conceptofmind:
> > Hi,
> > Pre-processing steps need to be applied to the HTML/XML before text extraction. CourtListener contains malformed HTML/XML that requires upstreamed content from CAP.
> > BS4 uses its own LXML backend extraction which is typically worse quality than Trafilatura extraction when precision is set True. I have already processed all of the data and uploaded the newest dump. I am waiting on HF to remove my rate limits so that I can finish it.
> 
> What kind of preprocessing steps are needed? What upstream content from CAP does it require? The current PR does not show these. 

[In the issue](https://github.com/r-three/common-pile/issues/21#issuecomment-2146377876), I stated a separate PR is required to handle additional processing. This requires the new CAP data dumps to have the opinionated fixes by CL applied and streamlined into the bulk CL data dump. Jack Cushman, Kevin Ramirez, Mike Lissner, and I are working on the most efficient way to provide this set of fixes to the community.

> I just applied BS4 text extraction on the raw data and it works great. If Trafilatura works better, could you include it and all the preprocessing you mentioned in the CR please? 

You can review the precision and F1 of different text extractors here:
```
| Model         | Mean Precision | Mean F1 | Median Precision | Median F1 |
|---------------|----------------|---------|------------------|-----------|
| Trafilatura   | 0.913          | 0.883   | 0.989            | 0.957     |
| DOM Distiller | 0.894          | 0.858   | 0.983            | 0.959     |
| Web2Text      | 0.797          | 0.841   | 0.885            | 0.917     |
| Boilerpipe    | 0.908          | 0.834   | 0.973            | 0.946     |
| Dragnet       | 0.901          | 0.823   | 0.980            | 0.943     |
| BTE           | 0.796          | 0.817   | 0.927            | 0.936     |
| Newspaper3k   | 0.896          | 0.816   | 0.994            | 0.958     |
| news-please   | 0.895          | 0.815   | 0.994            | 0.958     |
| Goose3        | 0.899          | 0.810   | 0.999            | 0.940     |
| BoilerNet     | 0.840          | 0.798   | 0.944            | 0.895     |
| ExtractNet    | 0.858          | 0.791   | 0.963            | 0.911     |
| jusText       | 0.794          | 0.759   | 0.949            | 0.904     |
| lxml Cleaner  | 0.615          | 0.717   | 0.670            | 0.798     |
| html_text     | 0.567          | 0.683   | 0.506            | 0.667     |
| BS4           | 0.563          | 0.680   | 0.506            | 0.669     |
| inscriptis    | 0.557          | 0.673   | 0.483            | 0.649     |
| XPath Text    | 0.550          | 0.664   | 0.510            | 0.674     |
```
Trafilatura with precision set to True will have even better results than the above. As you can see, BS4 is ranked quite low. Irregardless of the results above if we want to be consistent with Dolma which is used throughout this project we should use Trafilatura. It is a single-line adjustment to the above code, `trafilatura.extract(filecontent, favor_precision=True)`.

> A stand-alone dataset won't be very helpful. Thanks.

:joy::joy::joy::joy::joy::joy::joy:

blester125:
How variable is the markup in each example for this dataset? Part of the point of splitting work by datasets was to have someone work with the data and figure out patterns in the way a particular source formats their data so we can create a *better* text extractor/clean up for *that* data source instead of just dumping it into an off the shelf tool that tries to handle everything

conceptofmind:
> How variable is the markup in each example for this dataset? Part of the point of splitting work by datasets was to have someone work with the data and figure out patterns in the way a particular source formats their data so we can create a _better_ text extractor/clean up for _that_ data source instead of just dumping it into an off the shelf tool that tries to handle everything

Variability between CAP and CL is currently in "xml_harvard" which contains a lot of useful information. These documents require opinionated fixes from places like here: https://github.com/freelawproject/courtlistener/blob/a3bc7b472a1736bc0cb0e92cfa096dec12960e17/cl/corpus_importer/management/commands/harvard_opinions.py#L286

Which can be seen in a merged set of json here: https://github.com/freelawproject/opinionated/tree/main/data/harvard

This needs to be upstreamed in the current CAP dumps to CL. We are working on this.

Malformed HTML/XML can be seen here: `<span class="star-pagination" number="848" pagescheme="<span class="citation" data-id="4498419">`. 

There are numerous different instances of these cases. Once the upstreamed dumps are merged then CAP said this issue should not occur. I have written individual regex to cover as many of these edge cases as I could find. There are additionally child nodes that are being removed in relation to the edge case above as well.

There are also places in which I am writing a regex to improve tag `'<jurissystem' -> '<div'` standardization for extraction. These need to be handled independently as if you remove the incorrect tags then applying the upstream fixes between CAP and CL will have issues.

Other than that it is cleaning out standard things like boilerplate and OCR errors.

conceptofmind:
Additional changes will need to be made to drop the Harvard XML column and replace it with the new fixed HTML column from the updated dumps here: https://huggingface.co/datasets/conceptofmind/CAP

This will require the two datasets to be merged after appropriate column ordering from CAP.

PR #84, opened 2024-06-10
Regulations

### LINKED ISSUE #32: Regulations.gov
Federal regulation proposals and user comments.

- [ ] Get list of documents via Regulations.gov API
- [ ] Download docs (bulk download or individually?)

nkandpa2:
This PR closes #32. The data from 2000-2023 for the agencies listed below has been collected and is on HF [here](https://huggingface.co/datasets/nkandpa2/regulations). The total amount of data is around 2B tokens and 1.3GB on disk.

Agencies:
- BIS
- DOT
- EPA
- FAA
- FDA
- FEMA
- FERC
- FMCSA
- FRA
- NHTSA
- OSHA
- PHMSA
- SEC
- USCG


craffel:
I believe that for most of the other sources I'm aware of, there are typically about 1/2 as many tokens as there are on-disk bytes of gzipped jsonl. In your case there are about 2x as many, any idea where this discrepancy comes from?

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
+"""Build index of document URLs from a Regulations.gov bulk download file"""
```
blester125, line 1: copy pasted docstring

regulations/build-index.py
```diff
@@ -0,0 +1,101 @@
...
+        if os.path.exists(output_file) and not args.overwrite:
+            logger.info(f"Skipping {agency} data from {year} -- existing output file")
+            continue
+
+        logger.info(f"Processing {agency} data from {year}")
+
+        index = defaultdict(list)
+        num_skipped = 0
+        num_parsed = 0
+        with open(input_file, "r") as f:
```
blester125, line 81: A file that is going to be used with the csv reader should be opened with `newline=''`

regulations/build-index.py
```diff
@@ -0,0 +1,101 @@
...
+import parsing
+from tqdm.auto import tqdm
+
+from licensed_pile import logs
+
+csv.field_size_limit(sys.maxsize)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser("Regulations.gov index builder")
```
blester125, line 20: Missing `description=...`, as is this overwrites the program name

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+
+import requests
+from bs4 import BeautifulSoup
+from tqdm.auto import tqdm
+
+from licensed_pile import logs
+
+
+def parse_args():
+    parser = argparse.ArgumentParser("Regulations.gov index builder")
```
blester125, line 18: Same as above

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+    parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        default=False,
+        help="Overwrite existing files in output directory",
+    )
+    return parser.parse_args()
+
+
+def convert_htm(input_path, output_path):
```
blester125, line 65: prefer naming this `convert_html`, `htm` is super legacy from when windows couldn't have an extension with > 3 chars lol

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+    try:
+        with open(input_path, "r") as f:
+            text = f.read()
+    except UnicodeDecodeError:
+        try:
+            with open(input_path, "r", encoding="windows-1252") as f:
+                text = f.read()
+        except:
```
blester125, line 73: Can we add logging here? An error log that this file wasn't readable?

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+            return
+    soup = BeautifulSoup(text, "html.parser")
+    pre_tag = soup.find("pre")
```
blester125, line 76: This is the best place to use the walrus operator!

```python
parsed_text = pre_tag.get_text() if (pre_tag := soup.find("pre")) else text
```

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+    with open(output_path, "w") as f:
+        f.write(parsed_text)
+
+
+def convert_doc(input_path, output_path):
+    with open(output_path, "w") as f:
+        try:
+            subprocess.run(["catdoc", input_path], stdout=f, timeout=60)
+        except subprocess.TimeoutExpired:
+            return
```
blester125, line 93: Can we add a similar log on error here?

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def convert_docx(input_path, output_path):
+    try:
+        subprocess.run(["docx2txt", input_path], timeout=60)
+    except subprocess.TimeoutExpired:
+        return
+    subprocess.run(["mv", f"{os.path.splitext(input_path)[0]}.txt", output_path])
```
blester125, line 101: prefer `os.rename` over a subprocess

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def convert_txt(input_path, output_path):
+    subprocess.run(["cp", input_path, output_path])
```
blester125, line 105: prefer `shutil.copyfile` over a subprocess

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+
+def convert_doc(input_path, output_path):
+    with open(output_path, "w") as f:
+        try:
+            subprocess.run(["catdoc", input_path], stdout=f, timeout=60)
+        except subprocess.TimeoutExpired:
+            return
+
+
+def convert_docx(input_path, output_path):
```
blester125, line 96: Can you add a bit of explanation to this? A docx file that times out can be assume to already be a txt file?

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+    try:
+        subprocess.run(["docx2txt", input_path], timeout=60)
```
blester125, line 98: Can you add a comment about why `doc2txt` doesn't need a stdout redirection? Like that it auto creates a new `.txt` file?

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+    logger = logs.get_logger("regulations")
+    convert_funcs = {
+        ".txt": convert_txt,
+        ".htm": convert_htm,
+        ".doc": convert_doc,
+        ".docx": convert_docx,
+    }
+
+    pbar = tqdm()
+    postfix = {"Skipped": 0}
```
blester125, line 118: You can replace this is a `collection.Counter` to avoid having to preinitialize all the file types, you can just call the `+=`s

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+            input_dir = os.path.join(args.input_dir, year, agency)
+            output_dir = os.path.join(args.output_dir, year, agency)
+            os.makedirs(output_dir, exist_ok=True)
+
+            for file in os.listdir(input_dir):
+                file_name, file_type = os.path.splitext(file)
+                input_file = os.path.join(input_dir, file)
+                output_file = os.path.join(output_dir, f"{file_name}.txt")
+
+                if os.path.exists(output_file) and not args.overwrite:
```
blester125, line 134: We should add a log statement that this file was skipped, info level probably

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+                    postfix["Skipped"] += 1
+                elif file_type in args.file_types:
+                    func = convert_funcs[file_type]
+                    func(input_file, output_file)
+                    postfix[file_type] += 1
+                else:
```
blester125, line 140: Probably log a warning here about a file of the wrong type being found

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+        ".doc": convert_doc,
+        ".docx": convert_docx,
+    }
+
+    pbar = tqdm()
+    postfix = {"Skipped": 0}
+    for file_type in args.file_types:
+        postfix[file_type] = 0
+
+    for year in args.years:
```
blester125, line 122: Is there a reason you didn't use the `(year, agency)` itertools product here that you used in the other script? If you did you could probably use that same `tqdm.auto` to avoid having the `pbar.update(1)` later

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+import os
+
+import requests
+from tqdm.auto import tqdm
+
+from licensed_pile import logs
+
+
+def parse_args():
+    parser = argparse.ArgumentParser("Regulations.gov index builder")
```
blester125, line 15: Same as above, description is wrong too

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument(
+        "--years",
+        nargs="+",
+        default=list(map(str, range(2000, 2024))),
+        help="Years of data to process",
+    )
+    parser.add_argument(
+        "--agencies",
+        nargs="+",
+        default=[
```
blester125, line 29: A read-only value like this should probably be a tuple instead of a list. There are a few other places that should have this same change

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+            os.makedirs(output_dir, exist_ok=True)
+
+            for file in os.listdir(input_dir):
+                file_name, file_type = os.path.splitext(file)
+                input_file = os.path.join(input_dir, file)
+                output_file = os.path.join(output_dir, f"{file_name}.txt")
+
+                if os.path.exists(output_file) and not args.overwrite:
+                    postfix["Skipped"] += 1
+                elif file_type in args.file_types:
```
blester125, line 136: We should probably convert `args.file_types` into a set at some point as there are a lot of `in` queries for it.

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        default=False,
+        help="Overwrite existing files in output directory",
+    )
+    return parser.parse_args()
+
+
+def download_file(url, filename):
```
blester125, line 62: Was there something the shared function in `scrapy.py` wasn't able to do?

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+        return response.status_code
+    else:
+        return -1
+
+
+def main(args):
+    logger = logs.get_logger("regulations")
+    for year in args.years:
+        for agency in args.agencies:
+            os.makedirs(os.path.join(args.output_dir, year, agency), exist_ok=True)
```
blester125, line 79: This could be the first line of the loop over `itertools.product` yeah? Seems simpler to do it there instead of having these extra for loops

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+                        continue
+                    if file["File Type"] not in args.file_types:
+                        pbar_stats["Wrong File Type"] += 1
+                        continue
+                    ret = download_file(file["URL"], output_file)
+                    if ret == 200:
+                        pbar_stats["Downloaded"] += 1
+                    else:
+                        pbar_stats["Errors"] += 1
+                    pbar.set_postfix(pbar_stats)
```
blester125, line 118: We should log all the different things that are checked for here, that way we can re-compute these stats from earlier runs based on the logs when the stdout has gone away.

regulations/convert.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def convert_txt(input_path, output_path):
+    subprocess.run(["cp", input_path, output_path])
+
+
+def main(args):
+    logger = logs.get_logger("regulations")
+    convert_funcs = {
+        ".txt": convert_txt,
```
blester125, line 111: We should add a comment that our file exts include the `.` as they often don't. I don't think it is worth changing, but it is something to be mindful of when manipulating paths

regulations/parsing.py
```diff
@@ -0,0 +1,192 @@
...
+    return Rule(
+        document_id=record["Document ID"],
+        posted_date=record["Posted Date"],
+        title=record["Title"],
+        topics=parse_topics(record["Topics"]),
+        content_files=parse_files(record["Content Files"]),
+    )
+
+
+def parse_proposed_rule(record: pd.Series):
```
blester125, line 124: Where are these functions used? It it like auto dispatched? I don't see references to them anywhere.

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+            index = json.load(f)
+
+        pbar_stats = {
+            "Downloaded": 0,
+            "Errors": 0,
+            "Already Exists": 0,
+            "Wrong File Type": 0,
+        }
+        pbar = tqdm(index.items())
+        for doc_id, metadatas in pbar:
```
blester125, line 101: Can you add a comment explaining this metadata format? It seems like each doc can have multiple metadata entries? And then each one is a different file type? Or can they overlap but we just use the first value (because the file existence check)?

regulations/to-dolma.py
```diff
@@ -0,0 +1,121 @@
+import argparse
```
blester125, line 1: Missing docstring

regulations/to-dolma.py
```diff
@@ -0,0 +1,121 @@
...
+    parser.add_argument("--workers", type=int, default=10, help="Number of threads")
+    args = parser.parse_args()
+    return args
+
+
+def generate_records(args):
+    for year, agency in itertools.product(args.years, args.agencies):
+        index_path = os.path.join(args.index_dir, year, f"{agency}.json")
+        if not os.path.exists(index_path):
+            continue
```
blester125, line 75: Probably add a warning log statement here if a full year of data is missing lol

regulations/to-dolma.py
```diff
@@ -0,0 +1,121 @@
...
+        with open(index_path, "r") as f:
+            index = json.load(f)
+
+        for doc_id, metadata_list in index.items():
+            for metadata in metadata_list:
+                file_path = os.path.join(args.file_dir, year, agency, f"{doc_id}.txt")
+                if not os.path.exists(file_path):
+                    continue
+                try:
```
blester125, line 84: Are their still files that need to opened with the `window-1252` encoding? I would have thought that the conversion functions would have turned everything to utf-8 default encoding?

regulations/to-dolma.py
```diff
@@ -0,0 +1,121 @@
...
+                    with open(file_path, "r") as f:
+                        text = f.read()
+                except UnicodeDecodeError:
+                    try:
+                        with open(file_path, "r", encoding="windows-1252") as f:
+                            text = f.read()
+                    except UnicodeDecodeError:
+                        continue
```
blester125, line 92: Probably add an error that this file couldn't be opened

regulations/to-dolma.py
```diff
@@ -0,0 +1,121 @@
...
+
+                url = None
+                for file_metadata in metadata["Content Files"]:
+                    if file_metadata["File Type"] in [".htm", ".txt", ".doc", ".docx"]:
+                        url = file_metadata["URL"]
```
blester125, line 97: I'm confused about this index format again, this is always going to set url to the last file type in the `metadata["content Files"]` field, but the downloads earlier seemed to download all the different formats?

regulations/to-dolma.py
```diff
@@ -0,0 +1,121 @@
...
+
+                record = {
```
blester125, line 99: Dolma format is incorrect, `document_type`, `title`, `agency` should all be in the `metadata` dict and `posted_date` should be `created`

regulations/download-files.py
```diff
@@ -0,0 +1,124 @@
...
+        return -1
+
+
+def main(args):
+    logger = logs.get_logger("regulations")
+    for year in args.years:
+        for agency in args.agencies:
+            os.makedirs(os.path.join(args.output_dir, year, agency), exist_ok=True)
+
+    for year, agency in itertools.product(args.years, args.agencies):
```
blester125, line 81: Should this downloading be parallelized at all? Or is the data small/you are parallelizing by running disjoint (year, agency) pairs on multiple workers?

regulations/README.md
```diff
@@ -0,0 +1,7 @@
+# Regulations.gov
+
+Proposed regulations and supporting documents published on [Regulations.gov](http://www.regulations.gov). Since each of these documents are authored by the US Federal Government, they are in the Public Domain.
+
+# Data Collection
+
+This collection's metadata was first gathered via a number of bulk download requests to Regulations.gov since each bulk download can only be for metadata related to a single agency over a single year span. In total, we requested metadata from 14 agencies between the years 2000 and 2023. To collect the documents, run the script `get-data.sh`. Internally this parses the metadata files to create an index of all file URLs referenced in the metadata. It then downloads all of the referenced .doc, .docx, .txt, and .htm files and converts each format to plaintext. Finally, it reads each of these converted files and stores them in a Dolma dataset. The resulting dataset is written to `data/regulations/v0`.
```
blester125, line 7: The mentioned `get-data.sh` script doesn't seem to be checked in

blester125 (changes requested):
Made some comments, I think the index metadata format should be clarified, there are some places where it seems to get processed in different ways.

I also didn't look at the usgpo files at all, I assumed they covered by your other PR

PR #86, opened 2024-06-20
Tool for comparing dolma data after preprocessing.

blester125:
This adds a script I have been using to spot check my text cleaning implementations. Point it at the aligned dolma shards from two different versions (raw vs v0, etc.) and you can use the buttons to move through example or jump to a specific example. Note: If documents are removed between the versions, this tool will be unaligned, could be fixed in the future by moving though the example id's instead of a simple numerical index.

The data is shown in independently scrollable boxes to help align sections when there are heavy changes.

Example Screenshot

image

### DIFF
--- a/licensed_pile/scripts/README.md
+++ b/licensed_pile/scripts/README.md
@@ -1,0 +1,28 @@
+# Scripts
+
+## Stats
+
+This script can be used with `size-stats-dolma` once the common-pile library is installed. It takes an `--input` that points to some dolma formatted data (a file, a glob, a dir) and outputs statistics about the size of the data.
+
+Example:
+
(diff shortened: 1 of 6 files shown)


licensed_pile/compare_data.py
```diff
@@ -0,0 +1,101 @@
+"""Compare pre and post processed dolma examples."""
```
craffel, line 1: More accurate would be "compare dolma examples across different processing stages"?

craffel (commented):
Nice! Maybe a stretch goal, but perhaps it would be cool to do some kind of highlighted diff between the two verisons?

PR #90, opened 2024-09-09
'Update DPI'

conceptofmind:
@craffel @shayne-longpre 

Please review the changes to DPI.

The changes to the downloader were made to bypass the Hugging Face dataset bug which caused issues with caching.

Should add Ahmad Anis as a contributor to the DPI section as he has helped resolve underlying issues with the dataset.

This still misses OIG but we are currently investigating what could be the potential issue.

Token Count:
```
Found 13 files to process
characters: 6.96Gc [00:12, 573Mc/s]
bytes: 7.12Gb [00:12, 586Mb/s]d/s] 
tokens: 1.15Gt [00:12, 95.1Mt/s]
documents: 9.69Md [00:12, 798kd/s]
shards: 13.0s [00:12, 1.07s/s]Mc/s] 
```

Thank you,

Enrico


shayne-longpre:
@conceptofmind I think it looks good, but can we add a check now in the `to-dolma.py` file that asserts every `ex['dataset']` is in the include.csv list of datasets? Thanks!

conceptofmind:
Will add

conceptofmind:
Added an assert for valid ids in include csv

data_provenance/to-dolma.py
```diff
@@ -132,11 +140,12 @@ def file_to_dolma(path: str, include_df: str, source_name: str = SOURCE_NAME):
                     "license_url": license_urls,
                     "language": langs,
                     "url": url,
-                    "dataset_id": ex["dataset"],
+                    "dataset_id": dataset_id,
                     "response": target_text,
                 },
             }
         )
+
```
shayne-longpre, line 148: Great! Also collect `seen_dataset_ids` then check here:

```
unseen_ids = valid_ids - set(seen_dataset_ids)
print(f"Unseen Datasets: {len(unseen_ids)} | {unseen_ids}")
```

shayne-longpre (commented):
LGTM after minor comment. Let me know that it passes asserts and what it prints out at the end?

data_provenance/to-dolma.py
conceptofmind: Will do

data_provenance/README.md
```diff
@@ -15,6 +15,6 @@ The specific filter settings are here: https://github.com/Data-Provenance-Initia
 
 Here is the process to download the data, from inside the `data_provenance` dir:
 
-1. Run `python download.py --include include.csv`
+1. Run `python3 hf_downloader.py`
```
craffel, line 18: I don't think we should specify python3 - we can assume that either users have python mapped to python3 or that they know they might need to do this (also to follow the convention with other readmes in this repo).

data_provenance/hf_downloader.py
```diff
@@ -0,0 +1,80 @@
...
+        local_dir=args.top_dir,
+        allow_patterns="*.jsonl",
+        token=args.hf_token,
+    )
+
+    logger.info(f"Saving {args.hf_dataset}")
+
+    hugging_dir = os.path.join(args.top_dir, ".huggingface")
+    if os.path.exists(hugging_dir):
+        shutil.rmtree(hugging_dir)
```
craffel, line 51: I guess if we are going to be nuking this path we should be careful here, right? I.e. if hugging_dir exists when the script is called, we should exit to avoid nuking an existing huggingface dir?

craffel (changes requested):

data_provenance/hf_downloader.py
conceptofmind: This is currently there to get around the issue when a previously cached dataset exists in that path then it will be nuked otherwise will throw an error from datasets. 

Can think about adding some edge cases there.

data_provenance/README.md
conceptofmind: Will do

data_provenance/to-dolma.py
conceptofmind: Or maybe if no assert used:

```python
if dataset_id in valid_ids:
  seen_dataset_ids.add(dataset_id)
else:
  logger.warning(f"Dataset ID '{dataset_id}' not found in include.csv")
```

shayne-longpre: @conceptofmind wait, this looks like a bug of some sort. There are 273 datasets (including many simple ones like dolly, helpsteer,  and many fc-flan/sni ones that are all omitted). I thought the new HF object had all of these?

conceptofmind: Looks like it is just a logging issue. Will be moved outside the loop

shayne-longpre: LGTM! Thanks for investigatin!

conceptofmind: Should be resolved. Can reopen if not

data_provenance/hf_downloader.py
craffel: My point is that if a user does something like specify ~ for --top_dir it would nuke their standard hugging face cache directory, which would be a pretty surprising result. I was just suggesting that you check whether --top_dir/.huggingface exists before putting stuff in it and then nuking it, because presumably --top_dir is supposed to be a dir where there's no .huggingface directory.

conceptofmind: Sounds good. Will add something for handling those cases.

conceptofmind: Something like
```python
hugging_dir = os.path.join(args.top_dir, ".huggingface")
if os.path.exists(hugging_dir):
    logger.error(f"The specified directory '{args.top_dir}' already contains a .huggingface folder. "
                 "Choose a different directory to avoid potential data loss.")
    return

os.makedirs(args.top_dir, exist_ok=True)
```

conceptofmind: and
```python
if top_dir == os.path.expanduser("~/.cache/huggingface"):
    logger.error(f"The specified directory '{top_dir}' is the default Hugging Face cache directory. "
                 "Choose a different directory to avoid data loss.")
    return
    ```

Comment (conceptofmind) [reply data_provenance/hf_downloader.py:51]:
```python
assert args.top_dir != os.path.expanduser("~/.cache/huggingface"), \
    f"The specified directory '{args.top_dir}' is the default Hugging Face cache directory. " \
    "Choose a different directory to avoid data loss."

hugging_dir = os.path.join(args.top_dir, ".huggingface")
assert not os.path.exists(hugging_dir), \
    f"The specified directory '{args.top_dir}' already contains a .huggingface folder. " \
    "Choose a different directory to avoid potential data loss."
    ```

Comment (conceptofmind) [reply data_provenance/hf_downloader.py:61]: Add some asserts to prevent hf cache removal

**Comment by craffel** [comment]
LGTM, adding @blester125 to do a final pass

**Review by craffel — APPROVED** [review]

File: data_provenance/hf_downloader.py [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+                name, ext = os.path.splitext(file)
+                dest_path = os.path.join(args.top_dir, f"{name}_{ext}")
+
+            with open(source_path, "rb") as f_in:
+                with gzip.open(dest_path + ".gz", "wb") as f_out:
+                    shutil.copyfileobj(f_in, f_out)
+
+            os.remove(source_path)
+
+        if root != args.top_dir:
```
blester125, line 78: If we pass `topdown=False` to `os.walk`, this remove dir should work more often, you won't be trying to remove a dir that has subdirs in it.

data_provenance/hf_downloader.py
```diff
@@ -0,0 +1,90 @@
...
+        allow_patterns="*.jsonl",
+        token=args.hf_token,
+    )
+
+    logger.info(f"Saving {args.hf_dataset}")
+
+    if os.path.exists(hugging_dir):
+        shutil.rmtree(hugging_dir)
+
+    for root, dirs, files in os.walk(args.top_dir):
```
blester125, line 63: Can we add some comments about what we are trying to do here? Its a lot to follow?

Not sure if these are 100% accurate 
```python
# Move all nested files to the top level
for file in files:
  ...
  # Update the path name is the file already exists
  while os.path.exists(...
  
  # gzip each file
  with open(...
```

data_provenance/hf_downloader.py
```diff
@@ -0,0 +1,90 @@
...
+        for file in files:
+            source_path = os.path.join(root, file)
+            dest_path = os.path.join(args.top_dir, file)
+
+            while os.path.exists(dest_path) or os.path.exists(dest_path + ".gz"):
```
blester125, line 68: What are the cases where there file name conflicts? Is there a way to avoid them?

data_provenance/hf_downloader.py
```diff
@@ -0,0 +1,90 @@
...
+                name, ext = os.path.splitext(file)
+                dest_path = os.path.join(args.top_dir, f"{name}_{ext}")
+
+            with open(source_path, "rb") as f_in:
```
blester125, line 72: Should there be a check that the file isn't already gzipped? The `while` loop above makes it seem like there can be gzip versions already floating around?

data_provenance/to-dolma.py
```diff
@@ -66,6 +66,22 @@
 SOURCE_NAME = "Data Provenance Initiative"
 
 
+def get_unique_ids(folder_path):
```
blester125, line 69: Probably add a comment that this assumes that there aren't subdirectories, or use a glob that supports finding nested jsonl.gz files.

data_provenance/to-dolma.py
```diff
@@ -151,6 +178,12 @@ def main(args):
     )
     to_dolma(examples, args.outdir, args.filename, args.shard_size)
 
+    valid_ids = set(include_df["Dataset ID"])
+    unique_doc_ids = get_unique_ids(args.outdir)
+    unseen_ids = valid_ids - unique_doc_ids
+    print(f"Unseen Datasets: {len(unseen_ids)} | {unseen_ids}")
```
blester125, line 184: prefer outputting via logging

data_provenance/to-dolma.py
```diff
@@ -151,6 +178,12 @@ def main(args):
     )
     to_dolma(examples, args.outdir, args.filename, args.shard_size)
 
+    valid_ids = set(include_df["Dataset ID"])
+    unique_doc_ids = get_unique_ids(args.outdir)
```
blester125, line 182: This function has a confusing name, dolma ids are example level, but you're actually using dataset level ids to make sure each dataset is processed? We should make that clearer

data_provenance/to-dolma.py
```diff
@@ -66,6 +66,22 @@
...
+    unique_doc_ids = set()
+
+    jsonl_files = [f for f in os.listdir(folder_path) if f.endswith(".jsonl.gz")]
+
+    for filename in tqdm(jsonl_files, desc="Processing files"):
+        file_path = os.path.join(folder_path, filename)
+        with gzip.open(file_path, "rt") as f:
+            for line in f:
+                data = json.loads(line)
+                doc_id = data.get("id")
```
blester125, line 79: Each dolma example from this pipeline will have a `.["meta"]["dataset_id"]` key. Can we use that instead of munging the example id?

blester125 (changes requested):
Mostly great! Thanks for fixing the issues! Just a few small tweaks

conceptofmind:
Will make all these changes when back later

data_provenance/hf_downloader.py
conceptofmind: I think only if the script were to crash/fail then the filename would conflict. Maybe make sure the directory is cleared out first?

data_provenance/to-dolma.py
conceptofmind: Made changes to use logging instead

data_provenance/to-dolma.py
conceptofmind: Adjusted names to be more clear that they are DPI related

data_provenance/to-dolma.py
conceptofmind: Rerunning this now to make sure looks alright

data_provenance/hf_downloader.py
conceptofmind: Added a check to see if file is already gzip

data_provenance/hf_downloader.py
conceptofmind: I agree this is def confusing. Tried to add comments to make more clear

data_provenance/hf_downloader.py
conceptofmind: Added

data_provenance/hf_downloader.py
```diff
@@ -0,0 +1,90 @@
...
+    logger = get_logger()
+    logger.info(f"Filtering to just the datasets in {args.hf_dataset}")
+
+    assert args.top_dir not in ["~/.cache/huggingface", "~"], (
+        f"Directory '{args.top_dir}' is the default Hugging Face cache directory. "
+        "Choose a different directory to avoid potential data loss."
+    )
+
+    hugging_dir = os.path.join(args.top_dir, ".huggingface")
+    assert not os.path.exists(hugging_dir), (
```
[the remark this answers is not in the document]
conceptofmind: Changed

data_provenance/hf_downloader.py
```diff
@@ -0,0 +1,90 @@
...
+        "--top_dir", default="data/raw-data-provenance", help="Path to output directory"
+    )
+    return parser.parse_args()
+
+
+def main(args):
+    logger = get_logger()
+    logger.info(f"Filtering to just the datasets in {args.hf_dataset}")
+
+    assert args.top_dir not in ["~/.cache/huggingface", "~"], (
```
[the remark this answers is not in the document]
conceptofmind: Changed

data_provenance/to-dolma.py
conceptofmind: Added glob support and rerunning now

conceptofmind: Reran and works

data_provenance/to-dolma.py
conceptofmind: Reran and works. got 315

conceptofmind:
@blester125 Just following up on this? Does it look good to go?

shayne-longpre (approved):

blester125 (approved):

shayne-longpre:
@blester125 @conceptofmind @craffel can this be merged?

craffel:
Whoops, thought we merged it already, thanks.

PR #91, opened 2024-09-20
Add domain counting script

wildphoton:
Script for extracting the sub-urls and counting their frequency and total-text-length in CCCC dataset.

### DIFF
--- a/dolma-cccc/domain_counter.py
+++ b/dolma-cccc/domain_counter.py
@@ -1,0 +1,239 @@
+from huggingface_hub import snapshot_download
+
+import os
+import gzip
+import json
+from urllib.parse import urlparse
+from collections import defaultdict
+from concurrent.futures import ProcessPoolExecutor, as_completed, wait, FIRST_EXCEPTION
+from tqdm import tqdm
+import gc
+import signal
+import sys
+from datasets import Dataset, load_dataset
+
+
+def process_file(file_path):
+    suburl_counter = defaultdict(int)
+    text_length_counter = defaultdict(int)  # Additional counter based on text length
+
+    # First, count the total number of lines in the file for progress bar initialization
+    # with gzip.open(file_path, 'rt', encoding='utf-8') as f:
+    #     total_lines = sum(1 for _ in f)
+
+    # Re-open the file and process with progress bar
+    with gzip.open(file_path, 'rt', encoding='utf-8') as f:
+        with tqdm(desc=f"Processing {os.path.basename(file_path)}", leave=False,
+                  position=1) as line_progress:
+            for line in f:
+                try:
+                    data = json.loads(line)
+                    url = data.get('metadata', {}).get('warc_url', '')
+                    text = data.get('text', '')
+                    text_length = len(text)
+
+                    if url:
+                        suburls = extract_suburls(url)
+                        for suburl in suburls:
+                            suburl_counter[suburl] += 1
+                            text_length_counter[suburl] += text_length  # Increment based on text length
+
+                    line_progress.update(1)  # Update progress bar for each processed line
+                except json.JSONDecodeError:
+                    continue
+
+    gc.collect()  # Explicitly trigger garbage collection to free memory
+    return suburl_counter, text_length_counter
+
+
+def extract_suburls(url):
+    parsed_url = urlparse(url)
+    domain = parsed_url.netloc
+    path_parts = parsed_url.path.strip('/').split('/')
+
+    suburls = [domain, ]
+    current_path = domain
+
+    for part in path_parts[:-1]:
+        current_path = f"{current_path}/{part}"
+        suburls.append(current_path)
+
+    return suburls
+
+
+def merge_counters(counter_list):
+    merged_counter = defaultdict(int)
+    for counter in counter_list:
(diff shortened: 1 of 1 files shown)


dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+
+def extract_suburls(url):
+    parsed_url = urlparse(url)
+    domain = parsed_url.netloc
+    path_parts = parsed_url.path.strip('/').split('/')
+
+    suburls = [domain, ]
+    current_path = domain
+
+    for part in path_parts[:-1]:
```
blester125, line 57: Why are you indexing with `[:-1]` here? In the `path_parts` line you have `.strip('/')` so it seems like this function is ok with the final part of the path being a directory (and therefore should be part of this suburl list) but it gets filtered out?

dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+    def signal_handler(sig, frame):
+        print("Interrupt received, terminating...")
+        executor.shutdown(wait=False)
+        sys.exit(1)
+
+    # Register the signal handler
+    signal.signal(signal.SIGINT, signal_handler)
+
+    with tqdm(total=len(all_files), desc="Total files", position=0) as file_progress:
+        with ProcessPoolExecutor(max_workers=40) as executor:  # Adjust max_workers based on your system
```
blester125, line 89: This max worker should be a cli argument if you want people to adjust it themselves

dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+
+            except KeyboardInterrupt:
+                print("Ctrl+C pressed, terminating...")
+                executor.shutdown(wait=False)
+                sys.exit(1)
+
+    final_suburl_counter = merge_counters(suburl_results)
+    final_text_length_counter = merge_counters(text_length_results)
+
+    sorted_suburls = sorted(final_suburl_counter.items(), key=lambda item: item[1], reverse=True)
```
blester125, line 116: You're using a counter object, you can do `final_suburl_counter.most_common(10)` to get the top 10 sub urls by frequency.

dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+    print("Top 10 SubURLs by frequency:")
+    for suburl, count in sorted_suburls[:10]:
+        print(f"{suburl}: {count}")
+
+    sorted_text_lengths = sorted(final_text_length_counter.items(), key=lambda item: item[1], reverse=True)
```
blester125, line 121: Same as above

blester125, line 121: Something like

```
import operator as op

... = sorted(..., key=op.itemgetter(1), ...)
```

is generally preferred over a lambda to get a item like this

dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+    urls_to_remove = set()
+
+    # Iterate over each URL and compare it with its parent URL
+    for url, data in tqdm(url_dict.items()):
+        parent_url = find_parent_url(url)
+        if parent_url and parent_url in url_dict:
+            parent_data = url_dict[parent_url]
+            if (data['total_sample_count'] == parent_data['total_sample_count'] and
+                    data['total_text_length'] == parent_data['total_text_length']):
+                # If counts match, mark the higher-level URL for removal
```
blester125, line 165: The term `higher-level` is confusing, it makes it seem like you would be removing the parent url.

dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+dataset_name = "wildphoton/dolma-cccc-suburl-stats"
+# Run the processing
+root_folder = './data/dolma-cccc/data'  # Update with your root folder path
+
+# # Define the dataset name (this should be unique within your Hugging Face namespace)
+# dataset_name = "wildphoton/dolma-cccc-suburl-stats-CC-MAIN-2024-18"
+# # Run the processing
+# root_folder = './data/dolma-cccc/data/CC-MAIN-2024-18'  # Update with your root folder path
+
+token = "YOUR_HF_TOKEN"  # Replace with you own hf token if you want to submit the dataset
```
blester125, line 188: This should be a cli argument.

dolma-cccc/domain_counter.py
```diff
@@ -0,0 +1,239 @@
...
+
+    print(f"{len(urls_to_remove)}/{len(url_dict)} URL to be removed")
+
+    # Filter out the redundant URLs
+    filtered_data = [row for row in dataset if row['suburl'] not in urls_to_remove]
+
+    return filtered_data
+
+# download the dolma-ccc dataset for analysis
+snapshot_download(repo_id="allenai/dolma-cccc", local_dir='data/dolma-cccc', repo_type="dataset")
```
blester125, line 176: This should all be wrapped into a `main` function

blester125 (changes requested):
There are linting error that need to be fixed too

craffel:
We don't need this currently so I'll close it out, but good to have it around for reference, thanks.

PR #92, opened 2024-09-20
CCCC - license filter

baberabb:
added a script to filter the cccc dataset using a txt file (attached) of the manually verified license URLs.


cccc/scripts/filter_hf_ds.py
```diff
@@ -0,0 +1,179 @@
...
+    current_path = domain
+
+    for part in path_parts[:-1]:
+        current_path = f"{current_path}/{part}"
+        suburls.append(current_path)
+
+    return suburls
+
+
+def filter_condition(item: dict, URLS: set) -> bool:
```
blester125, line 92: All uppercase should be used for global variables. I like that this is passed into the function, but it should be called `urls`

cccc/scripts/filter_hf_ds.py
```diff
@@ -0,0 +1,179 @@
...
+    """Filters the data to only include the URLs in the urls.txt file"""
+    return extract_suburls(item["metadata"]["warc_url"])[0] in URLS
```
blester125, line 94: You're only ever going to be using the domain based on this code (`suburls` is returned by `extract_suburls` and that is always started with the domain).

We should convert this to explicitly use a domain extraction function.

cccc/scripts/filter_hf_ds.py
```diff
@@ -0,0 +1,179 @@
...
+        handlers=[logging.StreamHandler()],
+    )
+
+
+def read_file_to_list(file_path):
+    with open(file_path, "r") as file:
+        return [line.strip() for line in file]
+
+
+def process_file(
```
blester125, line 37: These files seem like they should be processable with dolma (they seem to be just be gzipped jsonl files.

We should move this to a dolma processor instead of hand rolling this processing and the parallelism. You should be able to subclass the `licensed_pile.write.ShardParallelProcessor`, override the `process_example` method to call the `filter_condition` function (return the example if it passes and `None` if it fails).

This should simplify things a lot.

cccc/scripts/urls_to_keep.txt
```diff
@@ -0,0 +1,538 @@
+column_1
```
blester125, line 1: Is this an error from this list being part of a csv at some point?

baberabb:
@blester125 Thanks for the comments! added the changes!

PR #95, opened 2024-10-30
Fix/wiki authors

blester125:
This PR adds the tool I used to fix the author formatting on wikiteam dumps and updates the dolma file creation code that caused the mismatch in the first place.

It also adds the tools I used to find and remove probable transcripts of copyrighted works.

### DIFF
--- a/wiki/README.md
+++ b/wiki/README.md
@@ -12,6 +12,13 @@ Insturctions for downloading and processing all the WikiMedia wikis can be found
 
 The Internet Archive WikiTeam scrapes many wikis across the net and publishes them to the Internet Archive. Of these many are openly licensed. The insturctions for downloading and processing those can be found in the `archive/` directory.
 
+#### Versions
+
+* `raw`: The original wikitext.
+* `v0`: Plain text from wtf\_wikipedia with special handling of math symbols.
+* `v1`: Removal of some errant html.
+* `v2`: Filtering of obvious license laundred content (transcripts).
+
 ### Wiki Scaping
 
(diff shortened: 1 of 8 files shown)


yzqzss:
> The wikiteam3 scraping tool, which is what most of the wiki's on the internet archive use, doesn't format page revision correctly. It creates the same xml format that the mediawiki format uses (`<page><revision>...</revision></page>`), but when there are multiple revisions to a single page it creates multiple `<page>...</page>` elements (one for each revision) instead of folding them into a single page with multiple revisions (`<page><revision>...</revision><revision>...</revision>...</page>`). We though about reducing these multiple edits into a single document, but we didn't as it would have been hard to tell which edits are useful to keep as full documents and which are small enough that they should be absorbed. Thresholding some kind of edit distance is basically what the (eventual) dedup process will be so it didn't seem worth it.

Hi, I'm wikiteam3 developer, let me explain this.

There are basically two xml dump strategies in wikiteam3:

1. Export page-by-page (per-page) like the MediaWiki cli tool does. Unfortunately, the 95-5 Rule also exists on wikis, and 95% of wiki pages have never been edited since they were created. Per-page export will waste a lot of request time (to prevent flood, we delay 1.5s between requests) on these unedited pages (each request can export <= 50 edits of a page). And in order to do the per-page export, we also need to request `API:Allpages` first to get all the page names of the entire wiki, which also takes time.
2. Use MediaWiki `API:Allrevisions` . By iterating over the revisions table, each request can steadily get 50~500 revisons (edits) on the entire wiki. This will produce the phenomenon you described (and if `pageid` of some revisions are the same in one API response, wikiteam3 will try to combine them into one <page>.). But the advantage is that we don't need to get the list of page names in advance, and each request can get more revisions.

The xmldump produced by `API:Allrevisions` also meets `MediaWiki's page export format`'s requirements.

PR #97, opened 2025-03-05
Remove problematic TSI datasets

shayne-longpre:


craffel:
I promise we will get to this during a post-NeurIPS code sprint :)

PR #99, opened 2025-03-25
Fix/stackexchange

blester125:
These are the changes I needed to make when reprocessing stack exchange now that there are community dumps and the official ones are missing data.


licensed_pile/write.py
```diff
@@ -120,9 +120,9 @@ def process_single(
                 update_interval = kwargs.pop("update_interval", 1)
                 debug = kwargs.pop("debug", False)
 
-                for i, line in enumerate(f):
-                    with logger(line=i):
-                        try:
+                try:
```
craffel, line 123: This catches exceptions that happen when reading lines from the file? I guess due to file corruption or something?

PR #102, opened 2025-03-27
feat/gharchive add progress bar

baberabb:


craffel:
Thanks for this, there are some magic constants in here though and some changes I'd want @blester125 to review so I will wait for his review.

blester125:
This isn't super important to include and would need a bunch of changes with the renaming so I'm thinking we just close it for now

PR #138, opened 2025-09-10
Fix helpsteer in DPI

shayne-longpre:
Remove as source as it contains synthetic data.


craffel:
Thanks Shayne. Did you fix this upstream too? Please fix the lints and we can merge.

PR #141, opened 2026-05-26
Fix arXiv LaTeX parsing edge cases and typos

### LINKED ISSUE #139: Possible bugs in Arxiv source
Thanks for making this codebase available!

I have two ideas on the arxiv pipeline, namely in `to-dolma.py`
- `os.path.splitext(info.name)[1] == ".tex"` only matches lowercase .tex, some authors use .TEX, so .lower() should be added here
- in `interpolate_document()`, it looks for `\input{}`. However, it is possible that this is commented out, in this case the code fragment would be inserted at the wrong place or twice. I suggest to do:
```python
pattern = r"^[^\S\n]*[^\S%]*?(?P<cmd>\\input{(?P<filename>.*?)})"

for m in re.finditer(pattern, contents, re.MULTILINE):
```

kaieberl:
Fixes a few edge cases in the arXiv LaTeX-to-Dolma conversion pipeline:

  - handle `.TEX` files case-insensitively
  - no interpolating for `\input{...}` / `\include{...}` commands that are
  commented out
  - resolve included TeX files case-insensitively and with omitted `.tex`
  extensions
  - fix the single-file gzip path to yield decoded contents instead of an
  undefined variable
  - replace deprecated `datetime.utcnow()`, still maintaining compatibility with Python 3.8
  - fix a few documentation/script typos, including the arXiv metadata
  filename and `bulk_download.py`

Closes #139.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ If you are looking for the data itself or our trained models, please see [our Hu
 ## Installation
 
 The majority of packages required for dataset creation can be installed with `pip install -r requirements.txt`.
-To make use of the shared functionality in the `common_pile` pckage, run `pip install -e .`.
-If you are on a system that doesn't support automatic installation of pandoc with `pypandoc_binary`, change it to `pypandoc` in the `requirements.txt` and and install pandoc manually.
+To make use of the shared functionality in the `common_pile` package, run `pip install -e .`.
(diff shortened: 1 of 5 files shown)


craffel:
@blester125 can you take a look? Thanks.

